PR #50, opened 2021-04-08
refactor(monitoring): standardize node data gathering

jhaeyaert:
Closes gravitee-io/issues/issues/5294


gravitee-node-api/src/main/java/io/gravitee/node/api/infos/NodeInfos.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public NodeStatus getStatus() {
+        return status;
+    }
+
+    public void setStatus(NodeStatus status) {
+        this.status = status;
+    }
+
+    public String getZone() {
```
brasseld, line 116: Zone is a concept specific to APIM, I'm not sure we need it as part of node module.
(And I'm not sure we need at all).

gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/NodeMonitoringService.java
```diff
@@ -0,0 +1,104 @@
...
+        if (repository == null) {
+            LOGGER.debug("There is nowhere to persist the monitoring data {}", monitoring);
+            return Single.just(monitoring);
+        }
+
+        final String monitoringKey = monitoring.getNodeId() + monitoring.getType();
+
+        Maybe<Monitoring> monitoringObs;
+
+        if (monitoringHolder.containsKey(monitoringKey)) {
```
brasseld, line 64: Can we add a small comment here just to explain why we are doing this (managing clustering, etc...)

PR #73, opened 2021-09-02
fix(event): propagate node stop event when shutdown

jhaeyaert:
Closes gravitee-io/issues#5966

### DIFF
--- a/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/monitor/NodeMonitorService.java
+++ b/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/monitor/NodeMonitorService.java
@@ -139,17 +139,8 @@ public class NodeMonitorService extends AbstractService {
   }
 
   @Override
-  protected void doStop() throws Exception {
+  public NodeMonitorService preStop() throws Exception {
     if (enabled) {
-      if (!executorService.isShutdown()) {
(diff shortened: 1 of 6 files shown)


gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/monitor/NodeMonitorService.java
```diff
@@ -161,6 +152,22 @@ protected void doStop() throws Exception {
...
+  }
+
+  @Override
+  protected void doStop() throws Exception {
+    if (enabled) {
+      if (!executorService.isShutdown()) {
+        LOGGER.info("Stop node monitor");
+        executorService.shutdownNow();
+      } else {
+        LOGGER.info("Gateway monitor already shut-downed");
```
brasseld, line 167: Gateway monitor -> Node monitor

brasseld (changes requested):

gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/monitor/NodeMonitorService.java
jhaeyaert: Good catch!

phiz71 (approved):

PR #81, opened 2021-09-30
Issues/#6292 reintroduce last evolutions to gravitee node vertx

kamiiiel:
Issues: https://github.com/gravitee-io/issues/issues/6292

### DIFF
--- a/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
+++ b/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
@@ -830,7 +830,7 @@ public class HttpServerConfiguration {
         );
 
       String sClientAuthMode = environment.getProperty(
-        "ssl.clientAuth",
+        prefix + "ssl.clientAuth",
         ClientAuth.NONE.name()
       );
       if (sClientAuthMode.equalsIgnoreCase(Boolean.TRUE.toString())) {
--- a/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
+++ b/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
@@ -17,14 +17,14 @@ package io.gravitee.node.vertx.configuration;
 
 import io.vertx.core.http.ClientAuth;
 import io.vertx.core.http.HttpServerOptions;
-import io.vertx.core.net.JksOptions;
-import io.vertx.core.net.PemKeyCertOptions;
-import io.vertx.core.net.PemTrustOptions;
-import io.vertx.core.net.PfxOptions;
+import io.vertx.core.net.*;
+import io.vertx.core.tracing.TracingPolicy;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
```diff
@@ -251,34 +289,104 @@ public HttpServerOptions getHttpServerOptions() {
...
+          );
+        }
+      } else if (
+        this.getKeyStoreType().equalsIgnoreCase(CERTIFICATE_FORMAT_SELF_SIGNED)
+      ) {
+        options.setPemKeyCertOptions(
+          SelfSignedCertificate.create().keyCertOptions()
+        );
+      }
       if (this.getKeyStorePath() != null) {
```
jhaeyaert, line 390: I don't understand why we configure the cert options another time 🧐

jhaeyaert (changes requested):
The .prettierrc file seems missing and should contains the following:
```yaml
printWidth: 140
tabWidth: 4
```

May I ask you to apply the prettier format because the code is a bit hard to read 🙏

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
```diff
@@ -654,6 +845,8 @@ public HttpServerConfiguration build() {
         environment.getProperty(prefix + "ssl.keystore.type", keyStoreType);
       this.keyStorePath =
         environment.getProperty(prefix + "ssl.keystore.path", keyStorePath);
+      this.keyStoreCertificates =
+        getCertificateValues("http.ssl.keystore.certificates");
```
jhaeyaert, line 849: Use prefix

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
```diff
@@ -664,14 +857,15 @@ public HttpServerConfiguration build() {
         environment.getProperty(prefix + "ssl.truststore.type", trustStoreType);
       this.trustStorePath =
         environment.getProperty(prefix + "ssl.truststore.path", trustStorePath);
+      this.trustStorePaths = getArrayValues("http.ssl.truststore.path");
```
jhaeyaert, line 860: Use prefix

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/configuration/HttpServerConfiguration.java
```diff
@@ -654,6 +845,8 @@ public HttpServerConfiguration build() {
         environment.getProperty(prefix + "ssl.keystore.type", keyStoreType);
```
jhaeyaert, line 845: What if I don't specify the `keystoreType` in my configuration ? I believe that the `keystoreType` will be valued to **`null`** instead of the default value (which was 'jks' if I'm not wrong).

We should make sure the default behavior remains the same as previously.

jhaeyaert: No you're right. I was misled by the default configuration in the gravitee.yaml file (https://github.com/gravitee-io/gravitee-gateway/blob/master/gravitee-gateway-standalone/gravitee-gateway-standalone-distribution/src/main/resources/config/gravitee.yml#L32-L33)

So, it means the original behavior is kept untouched 👍 🙂

PR #94, opened 2021-11-20
Remove Hazelcast in Standalone mode

kamiiiel:


gravitee-node-cluster/src/main/java/io/gravitee/node/cluster/standalone/StandaloneMap.java
```diff
@@ -33,9 +37,30 @@
...
+    Assert.notNull(name, "Name can't be null");
     this.name = name;
   }
 
+  @Override
+  public V put(K key, V value) {
+    Assert.notNull(key, "Key can't be null");
+    Assert.notNull(value, "Value can't be null");
+
+    mapListeners.forEach(
```
kamiiiel, line 49: Shall we run this on its own thread?

PR #96, opened 2021-12-02
feat: make gravitee propertysource enumerable

jgiovaresco:
In gravitee-common's EnvironmentUtils class, we have a getAllProperties that should return all properties of an environment.
Because Gravitee's property source was not enumerable nor composite, the helper function was not returning gravitee properties.
See https://github.com/gravitee-io/gravitee-common/blob/master/src/main/java/io/gravitee/common/util/EnvironmentUtils.java#L69

It will fix the [Cockpit initial plans](https://github.com/gravitee-io/gravitee-cockpit/blob/master/gravitee-cockpit-service/src/main/java/com/graviteesource/cockpit/service/spring/InitialPlansConfiguration.java#L21) loading where getting all environment properties were not returning properties from gravitee source and we will be able to update `gravitee-node` dependency: https://github.com/gravitee-io/gravitee-cockpit/pull/1452

This fix is to patch the master branch.


Okhelifi (approved):

jhaeyaert (approved):
LGTM 👍

PR #117, opened 2022-01-27
feat: addition of platform notifier mechanism

leleueri:
fixes gravitee-io/issues#6879

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationAcknowledgeRepository.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationAcknowledgeRepository.java
@@ -1,0 +1,44 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.api.notifier;
+
+import io.reactivex.Completable;
+import io.reactivex.Maybe;
+import io.reactivex.Single;
+
+/**
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface NotificationAcknowledgeRepository {
+  Maybe<NotificationAcknowledge> findById(String id);
+
+  Maybe<NotificationAcknowledge> findByResourceIdAndTypeAndAudienceId(
+    String resourceId,
+    String type,
+    String audienceId
+  );
+
+  Single<NotificationAcknowledge> create(
+    NotificationAcknowledge notificationAcknowledge
+  );
+
+  Single<NotificationAcknowledge> update(
+    NotificationAcknowledge notificationAcknowledge
+  );
+
+  Completable deleteByResourceId(String resourceId);
+}
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationDefinition.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationDefinition.java
@@ -1,0 +1,110 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.api.notifier;
+
+import java.util.Map;
+
+/**
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public class NotificationDefinition {
+
+  /**
+   * Resource concerned by the definition
+   */
+  private String resourceId;
+  /**
+   * Resource Type concerned by the definition
+   */
+  private String resourceType;
+  /**
+   * UserId targeted by the notification
+   */
+  private String audienceId;
+  /**
+   * Plugin Notifier configuration
+   */
+  private String configuration;
+  /**
+   * Type of notifier (plugin id)
+   */
+  private String type;
+  /**
+   * Cron expression used to evaluate the rule
+   */
+  private String cron;
+  /**
+   * Context used to evaluate the rule
+   */
+  private Map<String, Object> data;
+
+  public String getResourceId() {
+    return resourceId;
+  }
+
+  public void setResourceId(String resourceId) {
+    this.resourceId = resourceId;
+  }
+
+  public String getResourceType() {
+    return resourceType;
+  }
+
+  public void setResourceType(String resourceType) {
+    this.resourceType = resourceType;
+  }
+
+  public String getAudienceId() {
+    return audienceId;
+  }
+
+  public void setAudienceId(String audienceId) {
+    this.audienceId = audienceId;
+  }
+
+  public String getConfiguration() {
+    return configuration;
+  }
+
+  public void setConfiguration(String configuration) {
+    this.configuration = configuration;
+  }
+
+  public String getType() {
+    return type;
+  }
+
+  public void setType(String type) {
+    this.type = type;
+  }
+
+  public String getCron() {
+    return cron;
+  }
+
+  public void setCron(String cron) {
+    this.cron = cron;
+  }
+
+  public Map<String, Object> getData() {
+    return data;
+  }
+
+  public void setData(Map<String, Object> data) {
+    this.data = data;
+  }
+}
--- a/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
+++ b/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
@@ -1,0 +1,124 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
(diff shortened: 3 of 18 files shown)


gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationAcknowledgeRepository.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.api.notifier;
+
+import io.reactivex.Completable;
+import io.reactivex.Maybe;
+import io.reactivex.Single;
+import org.springframework.stereotype.Repository;
```
brasseld, line 21: to remove ?

gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationDefinition.java
```diff
@@ -0,0 +1,87 @@
...
+   */
+  private String type;
+  /**
+   * Cron expression used to evaluate the rule
+   */
+  private String cron;
+  /**
+   * Context used to evaluate the rule
+   */
+  private Map<String, Object> data = new HashMap<>();
```
brasseld, line 46: why is this instanciated and not final ?

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@Component
+public class NotifierServiceImpl implements NotifierService {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(
```
brasseld, line 42:
```suggestion
  private final Logger logger = LoggerFactory.getLogger(
```

gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationAcknowledgeRepository.java
```diff
@@ -0,0 +1,40 @@
...
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface NotificationAcknowledgeRepository {
+  Maybe<NotificationAcknowledge> findById(String id);
+
+  Maybe<NotificationAcknowledge> findByResourceIdAndAudienceId(
+    String resourceId,
+    String type,
+    String audience
```
jhaeyaert, line 32: Just for consistency: `audienceId`

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
```diff
@@ -0,0 +1,215 @@
...
+      this.notificationAcknowledgeRepository.findByResourceIdAndAudienceId(
+          this.definition.getResourceId(),
+          definition.getType(),
+          definition.getAudienceId()
+        )
+        .map(Optional::ofNullable)
+        .switchIfEmpty(Maybe.just(Optional.empty()))
+        .subscribe(
+          notifAck -> {
+            if (notifAck.isEmpty()) {
```
jhaeyaert, line 123: Here we never reschedule anything in case of error, is it normal ?
I saw also that, once the notification is successfully sent, no new scheduling is set. I think it makes sense but nothing inform the NotifierService that this trigger is not active anymore and could be removed from the triggers map.

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
```diff
@@ -0,0 +1,215 @@
...
+                            notificationAcknowledge.setType(
+                              definition.getType()
+                            );
+                            notificationAcknowledge.setAudienceId(
+                              definition.getAudienceId()
+                            );
+                            notificationAcknowledge.setResourceId(
+                              definition.getResourceId()
+                            );
+                            notificationAcknowledgeRepository
```
jhaeyaert, line 166: When multiple nodes will register the same notification trigger, there are good chances that the delay between the find ack and the create ack be big. The notification will be probably sent multiple times because of the cron (the process will be certainly executed at the exact same time) or because of the database replication delay.

Maybe something to discuss but a simple solution could be to apply a kind of random value when calculating the next tick ? (with something to explicitly activate when registering the trigger such as 'tryAvoidDuplicateNotification')

pom.xml
```diff
@@ -153,6 +154,20 @@
...
+                <artifactId>gravitee-plugin-notifier</artifactId>
+                <version>${gravitee-plugin.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>io.gravitee.el</groupId>
+                <artifactId>gravitee-expression-language</artifactId>
+                <version>${gravitee-expression-language.version}</version>
+                <scope>provided</scope>
+            </dependency>
```
jhaeyaert, line 169: Is this dependency really required ?

jhaeyaert (changes requested):

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
leleueri: why not simply rely on the database (mongo or RDBS) using a uniq index on the triplet (resource / type / aud )?

pom.xml
leleueri: indeeed

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+  private Vertx vertx;
+
+  @Autowired
+  private NotifierPluginFactory notifierFactory;
+
+  @Autowired
+  @Lazy
+  private NotificationAcknowledgeRepository notificationAcknowledgeRepository;
+
+  private Map<String, NotificationTrigger> triggers = new HashMap<>();
```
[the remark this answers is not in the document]
leleueri: indeed

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    String type,
+    String audience
+  ) {
+    return resource + "-" + type + "-" + audience;
+  }
+
+  @Override
+  public void unregister(String resourceId, String type, String audienceId) {
+    final String id = buildNotificationId(resourceId, type, audienceId);
+    Optional.ofNullable(triggers.get(id)).ifPresent(trigger -> trigger.stop());
```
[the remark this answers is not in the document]
leleueri: indeed

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/plugin/impl/NotifierPluginFactoryImpl.java
```diff
@@ -0,0 +1,267 @@
...
+        create0(
+          (Class<? extends Notifier>) ClassUtils.forName(
+            plugin.clazz(),
+            notifierClassLoader
+          ),
+          configuration
+        )
+      );
+    } catch (ClassNotFoundException e) {
+      e.printStackTrace();
```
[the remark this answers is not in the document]
leleueri: done

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
jhaeyaert: Because if you have multiple instances of mgt all instances will register and run the same trigger on its side with the same identifier resource / type / aud right ? So each instance will try to find if an ack exists in the DB in the exact same time and, if not, will send the same notification twice (or N times if you have N instances). You know what I mean ?

leleueri: ok, I define a random value (between 0 and 10 seconds). let me know if ok for you

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
```diff
@@ -0,0 +1,230 @@
...
+    } else {
+      LOGGER.debug("Notification Trigger can't be cancelled or doesn't exist");
+    }
+  }
+
+  private long computeNextAttempt() {
+    final LocalDateTime now = LocalDateTime.now();
+    final long delay =
+      toEpochMillis(this.cronExpression.next(now)) - toEpochMillis(now);
+    return randomDelayInSec > 0 ? delay + randomDelayInSec : delay;
```
jhaeyaert, line 110: Seems we mix millisecond and second here no ? 🧐

jhaeyaert (approved):
Looks good to me, just a point remaining regarding the mix between delay in second and delay in millisecond.

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
leleueri: indeed :mag:

gravitee-node-api/src/main/java/io/gravitee/node/api/notifier/NotificationAcknowledgeRepository.java
```diff
@@ -0,0 +1,44 @@
...
+import io.reactivex.Single;
+
+/**
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface NotificationAcknowledgeRepository {
+  Maybe<NotificationAcknowledge> findById(String id);
+
+  Maybe<NotificationAcknowledge> findByResourceIdAndAudienceId(
```
kamiiiel, line 29: Maybe It is better to add "AndType" to the name of the method

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/plugin/impl/NotifierPluginFactoryImpl.java
```diff
@@ -0,0 +1,267 @@
...
+        )
+      );
+    } catch (ClassNotFoundException e) {
+      logger.error("Unable to instantiate the class {}", plugin.clazz(), e);
+    }
+
+    return Optional.empty();
+  }
+
+  private <C extends NotifierConfiguration> Notifier create0(
```
kamiiiel, line 111: I think "0" at the end was just a mistake

tcompiegne (approved):

PR #125, opened 2022-03-01
fix: add rx vertx spring bean

jhaeyaert:


gaetanmaisse:
You will also maybe need another commit https://github.com/gravitee-io/gravitee-node/pull/124 1.19.x was broken without that 🤷🏻

gaetanmaisse (approved):

PR #129, opened 2022-03-14
fix(notifier): manage resourceType and provide a unregisterAll method…

leleueri:
… to remove trigger for a given resource at once

fixes gravitee-io/issues#7274

### DIFF
--- a/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
+++ b/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
@@ -16,6 +16,7 @@
 package io.gravitee.node.notifier;
 
 import static io.gravitee.node.notifier.NotifierUtils.buildNotificationId;
+import static io.gravitee.node.notifier.NotifierUtils.buildResourceKey;
 
 import com.google.common.collect.Maps;
 import io.gravitee.node.api.notifier.*;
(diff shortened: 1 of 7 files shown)


gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
```diff
@@ -75,19 +82,73 @@ public void register(
...
+    final String resourceKey = buildResourceKey(
+      definition.getResourceId(),
+      definition.getResourceType()
+    );
+    List<NotificationDefinition> allDefinitions = definitions.get(resourceKey);
+    if (Objects.isNull(allDefinitions)) {
+      allDefinitions = Collections.synchronizedList(new ArrayList<>());
+      definitions.put(resourceKey, allDefinitions);
+    }
+    allDefinitions.add(definition);
```
jhaeyaert, line 110: Maybe I'm wrong but I suspect a possible side effect if 2 `NotificationDefinition` on the same resource are registered almost in the same time (the `allDefinitions` list could be erased and one of the NotificationDefinition can be lost).

Do you think it is possible to rework it using a `computeIfAbsent` instead and rely on the concurrent map to ensure nothing could be lost ?

jhaeyaert (changes requested):
Only one thing to check regarding concurrency otherwise everything looks good to me.

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/NotifierServiceImpl.java
```diff
@@ -75,19 +82,76 @@ public void register(
...
+    List<NotificationDefinition> allDefinitions = definitions.get(resourceKey);
+    if (Objects.isNull(allDefinitions)) {
+      allDefinitions =
+        definitions.computeIfAbsent(
+          resourceKey,
+          key -> Collections.synchronizedList(new ArrayList<>())
+        );
+    }
+    allDefinitions.add(definition);
+  }
```
jhaeyaert, line 114: I think the problem remains the same no ?
What about this:

```java
definitions.computeIfAbsent(
          resourceKey,
          key -> Collections.synchronizedList(new ArrayList<>())
        ).add(definition);
```

leleueri: :face_exhaling: :+1:

jhaeyaert (approved):

PR #132, opened 2022-03-21
fix(vertx-node): use default values for array HttpServerConfiguraiton properties

mariano-fernandez:
## Issue

Similar to https://github.com/gravitee-io/gravitee-node/pull/131, default values are ignored for `HttpServerConfiguration` properties `keyStoreKubernetes` and `trustStorePaths`


brasseld:
Hey @mariano-fernandez 

Sounds like pretty straightforward. 
Still got a question regarding the branch. Why pushing on master instead of  1.20.X branch ?
Would have the same considerations regarding [straightforward](https://github.com/gravitee-io/gravitee-node/pull/131).

mariano-fernandez:
@brasseld you are right. I'll cherry-pick both commits and add them to 1.20.x in another PR.

should I do the same with 1.19.x?

PR #138, opened 2022-04-28
extend websocket config: frame , message size

benechiu:
**Issue**

If the websocket is used to transfer video streams, the default frame size / message size is too small which throws an exception.
Max frame length of 65536 has been exceeded

**Description**

Extend websocket frame size and message size in configuration.

Configuration:
http.websocket.maxWebSocketFrameSize
http.websocket.maxWebSocketMessageSize


brasseld:
Thanks for the contrib @benechiu

PR #149, opened 2022-08-03
feat: allows to include or exclude metrics labels

jhaeyaert:
**Issue**

https://github.com/gravitee-io/issues/issues/8218

**Description**

With that PR it is now possible to include or exclude a metric labels (aka tags) for a specific domain (aka category) by specifying it in the configuration.

By default, the following labels are included:

- `local`
- `http_method`
- `http_code`

On the apim gateway, if you want to include the tag `remote` only for the `http.client` (and avoid adding it for other categories, especially http.server), you can specify the following configuration: `gravitee.services.metrics.include.http.client[0]=remote`

You can also try to exclude the tag `local` on the `http.client` category because it is useless (vertx always set the value to '?'), you can add the following configuration: `gravitee.services.metrics.exclude.http.client[0]=local`

To check if the tags have been well taken in account, you can scrap the prometheus endpoint:

```bash
curl -u admin:<PASSWORD> http://localhost:18082/_node/metrics/prometheus
```

Note: don't forget to enable the metrics `gravitee.services.metrics.enabled=true`

---

 - use it directly by updating your project with version: `1.25.0-8218-prometheus-labels-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/1.25.0-8218-prometheus-labels-SNAPSHOT/gravitee-node-1.25.0-8218-prometheus-labels-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-api/pom.xml
+++ b/gravitee-node-api/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>io.gravitee.node</groupId>
         <artifactId>gravitee-node</artifactId>
-        <version>1.24.3</version>
+        <version>1.25.0-8218-prometheus-labels-SNAPSHOT</version>
     </parent>
 
     <artifactId>gravitee-node-api</artifactId>
--- a/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/VertxFactory.java
+++ b/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/VertxFactory.java
@@ -15,10 +15,15 @@
  */
 package io.gravitee.node.vertx;
 
+import static java.util.stream.Collectors.*;
+import static java.util.stream.StreamSupport.stream;
+
+import io.gravitee.common.util.EnvironmentUtils;
 import io.gravitee.node.api.Node;
 import io.gravitee.node.tracing.vertx.LazyVertxTracerFactory;
+import io.gravitee.node.vertx.metrics.ExcludeTagsFilter;
+import io.gravitee.node.vertx.metrics.RenameVertxFilter;
 import io.gravitee.node.vertx.verticle.factory.SpringVerticleFactory;
-import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.binder.jvm.ClassLoaderMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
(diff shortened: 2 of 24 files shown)


gravitee-node-api/pom.xml
```diff
@@ -24,7 +24,7 @@
     <parent>
         <groupId>io.gravitee.node</groupId>
         <artifactId>gravitee-node</artifactId>
-        <version>1.24.3</version>
+        <version>1.25.0-8218-prometheus-labels-SNAPSHOT</version>
```
ytvnr, line 27: note for later, should we apply the same principle as for apim ? i.e `<version>${revision}${sha1}${changelist}</version>`

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/VertxFactory.java
```diff
@@ -65,6 +67,10 @@ public class VertxFactory implements FactoryBean<Vertx> {
     @Autowired
     private LazyVertxTracerFactory vertxTracerFactory;
 
+    private Set<Label> micrometerLabels;
+
+    private Map<String, Set<Label>> micrometerExcludedLabelsByCategory;
```
ytvnr, line 72:
```suggestion
    private Set<Label> metricsLabels;

    private Map<String, Set<Label>> metricsExcludedLabelsByCategory;
```

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/VertxFactory.java
```diff
@@ -160,7 +164,74 @@ public boolean isSingleton() {
         return true;
     }
 
-    private Set<String> loadLabels() {
+    private void loadMetricLabels() {
+        final Map<String, Set<Label>> includedLabelsByCategory = readConfiguredLabelsByCategory("include");
+
+        // Include labels takes precedence over defined labels.
```
ytvnr, line 170: I really appreciate all those comments but I wonder if it's not preferable to have several little private function with their own intention. with time, it's easy to modify the code without modifying the comment.

Since your code is already well "split" by those comment it's easy to see dedicated methods
wdyt ?

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/metrics/ExcludeTagsFilter.java
```diff
@@ -0,0 +1,76 @@
...
+ * @author Jeoffrey HAEYAERT (jeoffrey.haeyaert at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public class ExcludeTagsFilter implements MeterFilter {
+
+    private final String category;
+
+    private final List<String> labels;
+
+    public ExcludeTagsFilter(String category, List<String> labels) {
```
ytvnr, line 39:
```suggestion
    public ExcludeTagsFilter(String category, List<String> excludedLabels) {
```

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/metrics/ExcludeTagsFilter.java
```diff
@@ -0,0 +1,76 @@
...
+                        return t;
+                    }
+
+                    filtered.set(true);
+                    return null;
+                })
+                .filter(Objects::nonNull)
+                .collect(toList());
+
+            if (filtered.get()) {
```
ytvnr, line 62: Maybe add a comment like `When some tags has been excluded, then replace tags collection`

gravitee-node-vertx/src/test/resources/logback.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ytvnr, line 1: usually this file is named `logback-test.xml` no?

ytvnr (commented):
non blocking comments

gravitee-node-api/pom.xml
jhaeyaert: Don't know what it is about but why not. Maybe something to add to the backlog.

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/VertxFactory.java
jhaeyaert: As discussed, after a second look, it seems already well split.

ytvnr (approved):

PR #166, opened 2023-02-10
fix(api): send head before writing response for /monitor

ytvnr:
**Issue**

https://gravitee.atlassian.net/browse/APIM-616

**Description**

For /_node/monitor, we tried to set status code after writing the response, but the head was already sent, causing an error


ytvnr:
close to target alpha

PR #199, opened 2023-03-28
fix: handle dry runs in upgraders

a-cordier:
---

 - use it directly by updating your project with version: `3.0.2-fix-dry-run-upgraders-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/3.0.2-fix-dry-run-upgraders-SNAPSHOT/gravitee-node-3.0.2-fix-dry-run-upgraders-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/upgrader/Upgrader.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/upgrader/Upgrader.java
@@ -21,6 +21,17 @@ package io.gravitee.node.api.upgrader;
  * @author GraviteeSource Team
  */
 public interface Upgrader {
+    /**
+     * Implementations can override this method to state that they ran without
+     * committing any change. If the return is true, the upgrade will not be stored
+     * and will run again on subsequent executions until the implementation returns false.
+     *
+     * @return true if the upgrader actually committed its changes, false otherwise.
+     */
+    default boolean isDryRun() {
+        return false;
+    }
+
     boolean upgrade();
 
     int getOrder();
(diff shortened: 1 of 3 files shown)


gravitee-node-api/src/main/java/io/gravitee/node/api/upgrader/Upgrader.java
```diff
@@ -21,6 +21,17 @@
...
  */
 public interface Upgrader {
+    /**
+     * Implementations can override this method to state that they ran without
+     * committing any change. If the return is true, the upgrade will not be stored
+     * and will run again on subsequent executions until the implementation returns false.
+     *
+     * @return true if the upgrader actually committed its changes, false otherwise.
+     */
+    default boolean isDryRun() {
```
guillaumelamirand, line 31: I am not sure to understand the feature. From the javadoc here, I understand the changes are not committed and no changes are made in the repository. From the code, I understand that only the UpgradeRecord is not committed to the database but any changes made by the upgrader are still performed. Could you explain ?

kamiiiel: All the changes made by the upgrader are in the dry run mode.
So it means nothing has been committed to the database by the upgrader actually.

guillaumelamirand: Is it up to the upgrader to manage that commitment?

kamiiiel: Yes, at least at the moment it is implemented like this. It can be improved later on. The concept on dry run is used only in APIM, btw.

PR #206, opened 2023-04-05
fix: prevent scheduleNextAttempt to happen if trigger is stopped

remisultan:
**Issue**

https://gravitee.atlassian.net/browse/COC-223

**Description**

This PR prevents the `scheduleNextAttempt` step to happen is the NotificationTrigger is already closed with a `started` flag.

**Additional context**

When calling `NotifierService.unregister` and subsequently `NotifierService.stopAndRemoveTrigger`, the `NotificationTrigger` can be in its `handle` phase. 

If we call `NotificationTrigger.stop` at the same time, a new timer will be created and there is no possibility of stopping the timer ever again due to the loss of reference of this trigger in the `NotifierService`.

This also creates a memory leak and a loss of performance since the `NotificationTrigger` keeps looping over until the node is restarted

---

 - use it directly by updating your project with version: `3.0.5-fix-notifier-rescheduled-when-stopped-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/3.0.5-fix-notifier-rescheduled-when-stopped-SNAPSHOT/gravitee-node-3.0.5-fix-notifier-rescheduled-when-stopped-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
+++ b/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
@@ -102,14 +102,12 @@ public class NotificationTrigger implements Handler<Long> {
     }
 
     public void stop() {
-        if (started.get() && this.scheduledTaskId != null) {
-            LOGGER.debug("Notification Trigger cancelled !");
+        started.set(false);
+        if (this.scheduledTaskId != null) {
             this.vertx.cancelTimer(this.scheduledTaskId);
-            started.set(false);
-            this.scheduledTaskId = null;
-        } else {
-            LOGGER.debug("Notification Trigger can't be cancelled or doesn't exist");
         }
+        this.scheduledTaskId = null;
+        LOGGER.debug("Notification Trigger cancelled !");
     }
 
     private long computeNextAttempt() {
--- a/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
+++ b/gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
@@ -33,6 +33,7 @@ import java.time.ZoneId;
 import java.util.Date;
 import java.util.Optional;
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.scheduling.support.CronExpression;
@@ -63,7 +64,7 @@ public class NotificationTrigger implements Handler<Long> {
 
     private Long scheduledTaskId;
 
-    private boolean started = false;
+    private final AtomicBoolean started;
 
     public NotificationTrigger(
         Vertx vertx,
@@ -81,6 +82,7 @@ public class NotificationTrigger implements Handler<Long> {
         this.condition = condition;
         this.resendCondition = resendCondition;
         this.cronExpression = CronExpression.parse(definition.getCron());
+        started = new AtomicBoolean();
         if (tryToAvoidMultipleNotif) {
             this.randomDelayInMs = Math.max(1, new Random().nextInt(10)) * 1000;
         } else {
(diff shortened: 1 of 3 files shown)


gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
```diff
@@ -186,10 +194,7 @@ public void handle(Long event) {
                                                             error
                                                         )
                                                     )
-                                                    .doFinally(() ->
-                                                        //trigger a new attempt
-                                                        this.scheduleNextAttempt()
-                                                    )
+                                                    .doFinally(this::scheduleNextAttempt)
```
jhaeyaert, line 197: I know that it's not this PR's main objective, but I can see here that, in case of an error, the next attempt will not be re-triggered. There is a doOnError that properly logs the error but the error is still propagated. Wouldn't an `onErrorResumeNext()` be more appropriate?

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
```diff
@@ -87,16 +89,21 @@ public NotificationTrigger(
...
     private void scheduleNextAttempt() {
-        this.scheduledTaskId = this.vertx.setTimer(computeNextAttempt(), this);
+        if (started) {
+            this.scheduledTaskId = this.vertx.setTimer(computeNextAttempt(), this);
+        }
     }
 
     public void stop() {
-        if (this.scheduledTaskId != null && this.vertx.cancelTimer(this.scheduledTaskId)) {
+        if (started && this.scheduledTaskId != null && this.vertx.cancelTimer(this.scheduledTaskId)) {
```
jhaeyaert, line 103: It probably requires synchronizing the started boolean so we ensure there is not a thread setting a timer while another is moving the `started` boolean to `false` (which will end in the same kind of situation)

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
```diff
@@ -186,10 +194,7 @@ public void handle(Long event) {
                                                             error
                                                         )
                                                     )
-                                                    .doFinally(() ->
-                                                        //trigger a new attempt
-                                                        this.scheduleNextAttempt()
-                                                    )
+                                                    .doFinally(this::scheduleNextAttempt)
                                                     .subscribe();
```
jhaeyaert, line 198: Probably out-of-scope here also but I can see several `subscribe()` calls here while I would expect a single `subscribe()` to the global chain.

jhaeyaert (changes requested):
Just one thing to potentially take care of. I also noticed 2 others things I let comment on.

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
guillaumelamirand: Atomicboolean would be cheaper don’t u think?

jhaeyaert: Totally agree 👍

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
remisultan: Thanks 👍 
Will look into it.

gravitee-node-notifier/src/main/java/io/gravitee/node/notifier/trigger/NotificationTrigger.java
```diff
@@ -87,16 +91,22 @@ public NotificationTrigger(
...
     private void scheduleNextAttempt() {
-        this.scheduledTaskId = this.vertx.setTimer(computeNextAttempt(), this);
+        if (started.get()) {
+            this.scheduledTaskId = this.vertx.setTimer(computeNextAttempt(), this);
+        }
     }
 
     public void stop() {
-        if (this.scheduledTaskId != null && this.vertx.cancelTimer(this.scheduledTaskId)) {
+        if (started.get() && this.scheduledTaskId != null) {
```
jhaeyaert, line 105: Hum... I have the feeling it doesn't solve the issue. I was thinking of using  `compareAndSet()` to ensure no concurrent actions.

jhaeyaert (changes requested):

jhaeyaert (approved):

PR #207, opened 2023-04-07
feat: add support for multi-servers

jhaeyaert:
**Issue**

https://gravitee.atlassian.net/browse/ARCHI-227

**Description**

The objective of this feature is to provide support for instantiating multiple servers. This is mandatory to support product features such as k8s Ingress for APIM (the ability to expose Ingresses on http or https both at the same time).

For that to be possible, the `VertxServerFactory` has been refactored and a `ServerManager` concept has been introduced to easily instantiate and get access to the created servers.

**Additional context**

---

 - use it directly by updating your project with version: `3.1.0-archi-227-multi-servers-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/3.1.0-archi-227-multi-servers-SNAPSHOT/gravitee-node-3.1.0-archi-227-multi-servers-SNAPSHOT.zip)


gravitee-node-api/src/main/java/io/gravitee/node/api/certificate/TrustStoreManager.java
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.api.certificate;
+
+/**
+ * @author Jeoffrey HAEYAERT (jeoffrey.haeyaert at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface TrustStoreManager {}
```
guillaumelamirand, line 22: Don't see any use of this interface? Even in VertxHttpServer, it returns null. Could you check on your side?

gravitee-node-api/src/main/java/io/gravitee/node/api/server/DefaultServerManager.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public void unregister(Server<?> server) {
+        servers.remove(server);
+    }
+
+    @Override
+    public List<Server<?>> servers() {
+        return servers;
```
guillaumelamirand, line 42: Shouln't we return a unmodifiable list to avoid others to remove/add without calling register/unregister method ?

gravitee-node-api/src/main/java/io/gravitee/node/api/server/Server.java
```diff
@@ -0,0 +1,33 @@
...
+package io.gravitee.node.api.server;
+
+import io.gravitee.node.api.certificate.KeyStoreManager;
+import io.gravitee.node.api.certificate.TrustStoreManager;
+
+/**
+ * @author Jeoffrey HAEYAERT (jeoffrey.haeyaert at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface Server<C extends ServerOptions> {
```
guillaumelamirand, line 25: Should we anticipate the need for a "type" such as HTTP, TCP etc ?

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/server/VertxServerFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final VertxHttpServerFactory httpServerFactory;
+
+    public VertxServerFactory(Vertx vertx) {
+        this.httpServerFactory = new VertxHttpServerFactory(vertx);
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public T create(C options) {
```
guillaumelamirand, line 37: Shouldn't we register every server created by the factory directly to the server manager; in order to always register newly created server ?

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/server/http/VertxHttpServer.java
```diff
@@ -0,0 +1,71 @@
...
+import lombok.RequiredArgsConstructor;
+
+/**
+ * @author Jeoffrey HAEYAERT (jeoffrey.haeyaert at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@RequiredArgsConstructor
+public class VertxHttpServer extends VertxServer<HttpServer, VertxHttpServerOptions> {
+
+    private final String id;
```
guillaumelamirand, line 35: You could move this to `VertxServer `

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/server/http/VertxHttpServer.java
```diff
@@ -0,0 +1,71 @@
...
+    private final Vertx vertx;
+    private final VertxHttpServerOptions options;
+    private final List<HttpServer> delegates = new CopyOnWriteArrayList<>();
```
guillaumelamirand, line 38: Could move this to `VertxServer `

guillaumelamirand (changes requested):

gravitee-node-api/src/main/java/io/gravitee/node/api/certificate/TrustStoreManager.java
jhaeyaert: Indeed, I've introduced these interfaces (`TrustStoreManager` and `KeyStoreManager`) just to make it clear that they will be stuck to each server in the future. I should probably remove them and introduce them later when the feature will be really developed 👍

gravitee-node-api/src/main/java/io/gravitee/node/api/server/Server.java
jhaeyaert: Indeed. I've added a `String type()`  method on the Server interface 👍

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/server/VertxServerFactory.java
jhaeyaert: I don't want to create a such link between the factory and the server manager. A server manager is not mandatory here and the factory could be used without it. I prefer to keep it like this for now.

guillaumelamirand: However, how will you keep track of every server instance?

jhaeyaert: It's up to the one that instantiates the server to keep track of it. Depending on the needs it can be done using a server manager or just keeping a single reference. For me, that's not mandatory to link `ServerManager` and `ServerFactory`.

guillaumelamirand (approved):

PR #217, opened 2023-04-28
[master] fix: license error in community mode

a-cordier:
see https://gravitee.atlassian.net/browse/APIM-343 
see https://github.com/gravitee-io/issues/issues/8659


gravitee-node-license/src/main/java/io/gravitee/node/license/LicenseService.java
```diff
@@ -219,6 +219,11 @@ private String getLicenseFile() {
     }
 
     private void verify() throws Exception {
+        if (license == null) {
+            logger.debug("License will not be verified as it has not been loaded");
+            return;
+        }
+
```
jhaeyaert, line 226: What could be the consequences when a valid license file is loaded and replaced with an invalid license file? Does the `license` attribute becomes `null` and could bypass the license check?

jhaeyaert (approved):
I'm good with the solution to avoid the NPE but could you look at the comment I left to double check there is no unwanted effect?

gravitee-node-license/src/main/java/io/gravitee/node/license/LicenseService.java
guillaumelamirand: Looks like the license is not hot reloaded right? So I assume if you update the file after the initial start, the license is not refreshed, isn't it ?

jhaeyaert: It appears there is a file watcher running in the background. That's why I'm asking myself if the fix could have unexpected consequences.

gaetanmaisse: Ok, @ThibaudAV and I checked the different cases and added a missing improvement to not break the file watcher when the license file is corrupted on https://github.com/gravitee-io/gravitee-node/pull/214

PR #234, opened 2023-05-25
fix: avoid npe when evaluating cpu load average

loriepisicchio:
**Issue**

https://gravitee.atlassian.net/browse/APIM-1566

**Description**

Avoid NPE on Windows


gcusnieux (approved):

loriepisicchio:
@gcusnieux I think it would generate too much logs, and apparently this is something that can happen on Windows...

PR #235, opened 2023-05-31
fix(cache): initialize the TTL value during the put action

leleueri:
**Description**

The TTL of the cache entry is never set, this PR initialize this value using the configuration object to evict the cache entry when it expires


leleueri:
Useless, 0 means infinite. a dedicated method is available to set TTL

PR #257, opened 2023-11-06
feat: add support for Gravitee Pem Registry

kamiiiel:
**Issue**

https://gravitee.atlassian.net/browse/APIM-2969

---

 - use it directly by updating your project with version: `2.1.0-gravitee-pem-registry-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/2.1.0-gravitee-pem-registry-SNAPSHOT/gravitee-node-2.1.0-gravitee-pem-registry-SNAPSHOT.zip)


gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/KubernetesSecretKeyStoreLoader.java
```diff
@@ -123,21 +122,43 @@ protected Flowable<Secret> watch() {
...
-                    options.getKeyStorePassword(),
-                    secret.getMetadata().getName()
-                );
+            keyStore = secretToKeyStore(secret);
         } else if (secret.getType().equals(KUBERNETES_OPAQUE_SECRET)) {
             if (options.getKeyStoreType().equalsIgnoreCase(CERTIFICATE_FORMAT_PEM)) {
                 return Completable.error(
                     new IllegalArgumentException("Pem format is not supported with opaque secret, use kubernetes tls secret instead.")
                 );
+            } else if (options.getKeyStoreType().equalsIgnoreCase(CERTIFICATE_FORMAT_PEM_REGISTRY)) {
```
benoitgravitee, line 134: Considering `http.ssl.keystore.kubernetes` is an array of secret references that end up being merged into a single keystore, I don't see the value of this feature. Looks like it comes from GKO project. Why not adding this capability to the helm chart to load a list of secrets rather than creating a Secret (that is not a secret) containing references to secrets. On top of that they cannot reside into another namespace than the 'registry secret' itself.

kamiiiel: The problem is that we don't know the secret names up-front and they will be added/updated/deleted in the run-time. So nothing is clear up front about the secretes. And yes, the pem-registry will be in the same namespace, since you can't refernce a secret outside the namespace for the ingress TLS

benoitgravitee: Ok, I get it. Why not in a config-map then ?

kamiiiel: It is not an issue, we can put it in a ConfigMap

kamiiiel: I changed the pem-registry to a ConfigMap

gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/KubernetesConfigMapKeyStoreLoader.java
```diff
@@ -117,41 +124,82 @@ protected Flowable<ConfigMap> watch() {
...
+                    .stream()
+                    .map(name -> {
+                        if (!StringUtils.hasLength(name)) {
+                            throw new IllegalArgumentException("Wrong or missing TLS secret name");
+                        }
+
+                        return kubernetesClient
+                            .get(ResourceQuery.secret(configMap.getMetadata().getNamespace(), name).build())
+                            .map(this::secretToKeyStore)
+                            .blockingGet();
```
jhaeyaert, line 146: You should avoid using `blockingGet()`.

gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/KubernetesConfigMapKeyStoreLoader.java
```diff
@@ -58,8 +59,11 @@ private void prepareLocations() {
...
             .forEach(location -> {
                 final Matcher matcher = CONFIGMAP_PATTERN.matcher(location);
+                final Matcher pemRegistryMatcher = PEM_REGISTRY_CONFIGMAP_PATTERN.matcher(location);
                 if (matcher.matches()) {
                     this.resources.put(matcher.group(1), ResourceQuery.<ConfigMap>from(location).build());
+                } else if (pemRegistryMatcher.matches()) {
+                    this.resources.put(location, ResourceQuery.<ConfigMap>from(location).build());
                 } else {
                     throw new IllegalArgumentException(
                         "You must specify a data when using configmap (ex: /my-namespace/configmaps/my-configmap/my-keystore)."
```
jhaeyaert, line 69: You should probably adapt the message.

gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/KubernetesConfigMapKeyStoreLoader.java
```diff
@@ -75,7 +79,10 @@ public static boolean canHandle(KeyStoreLoaderOptions options) {
             kubernetesLocations != null &&
             !kubernetesLocations.isEmpty() &&
             SUPPORTED_TYPES.contains(options.getKeyStoreType().toLowerCase()) &&
-            kubernetesLocations.stream().allMatch(location -> CONFIGMAP_PATTERN.matcher(location).matches())
+            (
+                kubernetesLocations.stream().allMatch(location -> CONFIGMAP_PATTERN.matcher(location).matches()) ||
+                kubernetesLocations.stream().allMatch(location -> PEM_REGISTRY_CONFIGMAP_PATTERN.matcher(location).matches())
```
jhaeyaert, line 84: Is it expected to avoid mixing configmap and pem registry configmap?
Should it be

```java
kubernetesLocations.stream().allMatch(location -> CONFIGMAP_PATTERN.matcher(location).matches() ||
       PEM_REGISTRY_CONFIGMAP_PATTERN.matcher(location).matches())
```

Instead?

gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/KubernetesConfigMapKeyStoreLoader.java
```diff
@@ -117,41 +124,82 @@ protected Flowable<ConfigMap> watch() {
...
-            .stream()
-            .filter(r ->
-                r.getNamespace().equalsIgnoreCase(configMap.getMetadata().getNamespace()) &&
-                r.getResource().equalsIgnoreCase(configMap.getMetadata().getName())
-            )
-            .findFirst();
-
-        if (optResource.isEmpty()) {
-            return Completable.error(new IllegalArgumentException("Unable to load keystore: unknown configmap."));
+        KeyStore keyStore = null;
```
jhaeyaert, line 127: Looks like this method becomes a bit large. Can you try to improve the readability(e.g. create private methods for each case)?

jhaeyaert (changes requested):

gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/KubernetesConfigMapKeyStoreLoader.java
jhaeyaert: You marked it as resolved but I don't see any change or comment so I unresolve it.

jhaeyaert (changes requested):
One remaining comment I've unresolved just to be sure it has been taken into account on your side.
Also, I'm not completely sure why this feature is about to be merged on the 2.0.x branch and not master (or alpha since there is a label mentioning it should be on alpha) 🤔

kamiiiel:
> One remaining comment I've unresolved just to be sure it has been taken into account on your side. Also, I'm not completely sure why this feature is about to be merged on the 2.0.x branch and not master (or alpha since there is a label mentioning it should be on alpha) 🤔

Because 2.0.x is used inside APIM 3.20 and we need to support that version as well.
The Alpha version will be used for the APIM 4.0, 4.1 and master

jhaeyaert (approved):

PR #286, opened 2024-02-07
fix: support keystore with mix of private key and cert entries

jhaeyaert:
**Issue**

https://gravitee.atlassian.net/browse/ARCHI-320

**Description**

When using a keystore containing a mix of private key entries and trusted certificate entries, the `KeyStoreLoaderManager` fails to load the entries.

This is because trusted cert entries are not password protected. Trying to get the entry with a password throws an exception with the following message: `trusted certificate entries are not password-protected`

---

 - use it directly by updating your project with version: `5.5.1-archi-320-fix-keystore-mix-entries-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/5.5.1-archi-320-fix-keystore-mix-entries-SNAPSHOT/gravitee-node-5.5.1-archi-320-fix-keystore-mix-entries-SNAPSHOT.zip)


gravitee-node-license/src/test/java/io/gravitee/node/license/DefaultLicenseManagerTest.java
```diff
@@ -11,6 +11,8 @@
 import io.gravitee.node.api.license.LicenseManager;
 import io.gravitee.plugin.core.api.PluginManifest;
 import io.gravitee.plugin.core.api.PluginRegistry;
+import io.reactivex.rxjava3.core.Flowable;
+import io.reactivex.rxjava3.processors.UnicastProcessor;
```
mariano-fernandez, line 15: Only new imports without new code. I'm pretty sure they aren't needed

mariano-fernandez (approved):
I tested the fix with Cockpit and everything looks good now

benoitgravitee (approved):

PR #296, opened 2024-03-06
feat: introduce vertx http and tcp client factories

ytvnr:
**Issue**

https://gravitee.atlassian.net/browse/APIM-3799

**Description**

Introduce `VertxHttpClientFactory` in a central place so that any project can use it.
In the case of APIM, it will be useful for
- http-proxy endpoint
- http health check api service (v4)
- http dynamic properties service (v4)

Also introduce `VertxTcpClientFactory`:
- useful for tcp-proxy endpoint

---

 - use it directly by updating your project with version: `5.9.0-apim-3799-extract-httpclient-factory-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/5.9.0-apim-3799-extract-httpclient-factory-SNAPSHOT/gravitee-node-5.9.0-apim-3799-extract-httpclient-factory-SNAPSHOT.zip)


gravitee-node-vertx/pom.xml
```diff
@@ -32,6 +32,7 @@
 
     <properties>
         <mockito-inline.version>3.12.4</mockito-inline.version>
+        <jackson-databind.version>2.16.1</jackson-databind.version>
```
jgiovaresco, line 35: Shouldn't we use the version coming from Gravitee BOM?

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/http/VertxHttpClientOptions.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * @author Yann TAVERNIER (yann.tavernier at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@Getter
+@Setter
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class VertxHttpClientOptions implements Serializable {
```
jgiovaresco, line 35: What about using `@Data` annotation to also define Equals/ToString. As it is a simple POJO defining an equals methods on the attributes could helps in tests when we want to check the construction of such object?

We could apply this to all new POJO added?

gravitee-node-vertx/src/test/java/io/gravitee/node/vertx/client/http/VertxHttpClientFactoryTest.java
```diff
@@ -0,0 +1,421 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/**
+ * @author Yann TAVERNIER (yann.tavernier at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@ExtendWith(MockitoExtension.class)
+@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
```
jgiovaresco, line 35: As you are not using _ in the test method name. This can be removed or change the method name to use snake case 🙈

jgiovaresco (approved):

remibaptistegio (approved):
I have just made a few comments, otherwise lgtm.

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/ssl/pkcs12/PKCS12KeyStore.java
```diff
@@ -0,0 +1,80 @@
...
+        } else if (getContent() != null && !getContent().isEmpty()) {
+            pfxOptions.setValue(io.vertx.core.buffer.Buffer.buffer(Base64.getDecoder().decode(getContent())));
+        } else {
+            throw new IllegalArgumentException("Missing PKCS12 keystore value for " + target);
+        }
+
+        pfxOptions.setAlias(getAlias());
+        pfxOptions.setAliasPassword(getKeyPassword());
+        pfxOptions.setPassword(getPassword());
+        clientOptions.setPfxKeyCertOptions(pfxOptions);
```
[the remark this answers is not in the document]
ytvnr: cool, I did not see that, I will simplify the method configureForClient to return the correct CertOptions implementation

remibaptistegio (approved):

PR #297, opened 2024-03-14
feat: a null license on create returns an oss license

jourdiw:
**Issue**

https://gravitee.atlassian.net/browse/APIM-4183

**Description**

When the base64 license is null, then an OSS license is returned.

Cockpit can delete an organization license, which effectively sends a payload with `license: null`. In this case, the `create` should yield an OSS license that APIM (Gateway + MAPI) can then use in `registerOrganizationLicense`.  

**Additional context**

---

 - use it directly by updating your project with version: `5.10.0-apim-4183-handle-null-org-license-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/5.10.0-apim-4183-handle-null-org-license-SNAPSHOT/gravitee-node-5.10.0-apim-4183-handle-null-org-license-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-license/src/main/java/io/gravitee/node/license/DefaultLicenseFactory.java
+++ b/gravitee-node-license/src/main/java/io/gravitee/node/license/DefaultLicenseFactory.java
@@ -47,15 +47,18 @@ public class DefaultLicenseFactory implements LicenseFactory {
     }
 
     @Override
-    public License create(@Nonnull String referenceType, @Nonnull String referenceId, @Nonnull String base64License)
+    public License create(@Nonnull String referenceType, @Nonnull String referenceId, String base64License)
         throws InvalidLicenseException, MalformedLicenseException {
-        return create(referenceType, referenceId, Base64.getDecoder().decode(base64License));
+        byte[] bytesLicense = Objects.isNull(base64License) ? null : Base64.getDecoder().decode(base64License);
+        return create(referenceType, referenceId, bytesLicense);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


gravitee-node-license/src/main/java/io/gravitee/node/license/DefaultLicenseFactory.java
```diff
@@ -47,8 +47,11 @@ public DefaultLicenseFactory(LicenseModelService licenseModelService) {
     }
 
     @Override
-    public License create(@Nonnull String referenceType, @Nonnull String referenceId, @Nonnull String base64License)
+    public License create(@Nonnull String referenceType, @Nonnull String referenceId, String base64License)
         throws InvalidLicenseException, MalformedLicenseException {
+        if (Objects.isNull(base64License)) {
+            return new OSSLicense(referenceType, referenceId);
+        }
```
jhaeyaert, line 54: Could be moved into the other create method taking byte array no? That way we cover both the case (nul String or byte[]). WDYT?

public License create(@Nonnull String referenceType, @Nonnull String referenceId, byte[] bytesLicense)

jhaeyaert (changes requested):

gravitee-node-license/src/main/java/io/gravitee/node/license/DefaultLicenseFactory.java
jourdiw: Yes I can do that 👍

jhaeyaert (approved):
LGTM 👍

PR #329, opened 2024-06-13
feat: add oas validation policy feature to license

ThibaudAV:
**Issue**
https://gravitee.atlassian.net/browse/APIM-4164

**Description**

Add oas validation policy feature to license.

**Additional context**

---

 - use it directly by updating your project with version: `5.16.0-APIM-4164-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/5.16.0-APIM-4164-SNAPSHOT/gravitee-node-5.16.0-APIM-4164-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-license/src/main/resources/license-model.yml
+++ b/gravitee-node-license/src/main/resources/license-model.yml
@@ -22,6 +22,7 @@ packs:
             - apim-policy-data-logging-masking
             - policy-data-logging-masking # for removal
             - apim-policy-geoip-filtering
+            - apim-policy-oas-validation
             - apim-policy-transform-avro-json
             - apim-policy-transform-protobuf-json
             - apim-policy-transform-avro-protobuf
--- a/gravitee-node-license/src/test/java/io/gravitee/node/license/DefaultLicenseFactoryTest.java
+++ b/gravitee-node-license/src/test/java/io/gravitee/node/license/DefaultLicenseFactoryTest.java
@@ -341,6 +341,7 @@ class DefaultLicenseFactoryTest {
             "apim-policy-interops-r-idp",
             "apim-policy-interops-a-sp",
             "apim-policy-interops-r-sp",
+            "apim-policy-oas-validation",
         };
         assertThat(license.getFeatures()).containsExactlyInAnyOrder(features);
 


jhaeyaert:
> 📝 There is also a resource associated with this policy. I think I use the same license key in this one. If it's the right thing to do ?¿

Which resource exactly? Is it a resource plugin?

ThibaudAV:
Yes `gravitee-resource-openapi-spec`

https://github.com/gravitee-io/gravitee-resource-openapi-spec
https://github.com/gravitee-io/gravitee-policy-oas-validation

jhaeyaert:
If I take some examples, for cache resource we have redis (oss) and in-memory (oss), for schema registry, we have confluent (ee). So keeping the same logic I would say that if gravitee-resource-openapi-spec is EE, then it requires to have de specific feature we can include into the license. wdyt?

Btw, I'm not sure what kind of resource is the gravitee-resource-openapi-spec and how many different implementations of such resources we plan to have?

ThibaudAV:
@jhaeyaert 
Well, that's why I'm confused. I can't find any obvious logic.
I haven't seen any other EE resources linked to a policy like this one. 

Schema registry for confluent have a dedicated feature `apim-en-schema-registry-provider` but all are pack into `event-native`.

if i add dedicated feature (ex: `apim-resource-validation-openapi-spec`) for this new resource i want to pack with the policy feature inside `enterprise-policy` 
moreover, in `enterprise-policy` we only have policies, but here we need the policy + resource pair.
😁 So I don't know. I have no preference and will follow your recommendation.

I think there will only be one implementation of this resource. 
it's more just there to "store" an OpenApi schema so that the policy can use it to validate stuff

jhaeyaert:
OK I believe we should have a sync call because I don't have the necessary knowledge of what the OAS resource does and how it works (the repo seems empty). Maybe this OAS resource is not really mandatory?

ThibaudAV:
Well after discussion the plans for the resource changed.
So this license feature is only there for the policy 
@jhaeyaert It's good for you ?

jhaeyaert (approved):

PR #339, opened 2024-06-20
fix: added one-second delay before repeat to avoid log flooding in ca…

lgw-gravitee:
continue of AM-3317. 
Added one second delay to avoid log flooding in case of invalid  kubernetes:// yaml property.

---

 - use it directly by updating your project with version: `5.18.2-AM-3317-fix-stackoverflow-ex-on-watch-loop-2-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/5.18.2-AM-3317-fix-stackoverflow-ex-on-watch-loop-2-SNAPSHOT/gravitee-node-5.18.2-AM-3317-fix-stackoverflow-ex-on-watch-loop-2-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-container/src/main/java/io/gravitee/node/container/spring/env/AbstractGraviteePropertySource.java
+++ b/gravitee-node-container/src/main/java/io/gravitee/node/container/spring/env/AbstractGraviteePropertySource.java
@@ -19,10 +19,13 @@ import io.gravitee.node.api.secrets.resolver.PropertyResolver;
 import io.gravitee.node.api.secrets.resolver.PropertyResolverFactoriesLoader;
 import io.gravitee.node.api.secrets.resolver.WatchablePropertyResolver;
 import io.reactivex.rxjava3.core.Flowable;
+import io.reactivex.rxjava3.core.Scheduler;
+import io.reactivex.rxjava3.functions.Supplier;
 import io.reactivex.rxjava3.schedulers.Schedulers;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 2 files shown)


gravitee-node-container/src/main/java/io/gravitee/node/container/spring/env/AbstractGraviteePropertySource.java
```diff
@@ -86,6 +87,7 @@ private void watchProperty(WatchablePropertyResolver<?> propertyResolver, String
             .defer(() -> propertyResolver.watch(value.toString()))
             .subscribeOn(Schedulers.io())
             .repeat()
+            .delay(1, TimeUnit.SECONDS)
```
jhaeyaert, line 90: Can you create a constant (ex: REPEAT_DELAY_SECONDS) or something?
Also, are you able to provide a test that validates the behavior 🙏 ?

jhaeyaert (changes requested):

gravitee-node-container/src/main/java/io/gravitee/node/container/spring/env/AbstractGraviteePropertySource.java
lgw-gravitee: I created a constant and I added a test... didn't have idea how to test more accurately

gravitee-node-container/src/test/java/io/gravitee/node/container/spring/env/WatchablePropertyResolverSubscriberTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Test
+    public void shouldRepeatFlowableWhenItsCompleted() throws Exception {
+        int secondsDelay = 1;
+        Supplier<Flowable<String>> flowableSupplier = () -> Flowable.just("1");
+        TestSubscriber<String> test = new AbstractGraviteePropertySource.FlowableRepeater<>(flowableSupplier, secondsDelay)
+            .asFlowable()
+            .test();
+
+        test.await(3, TimeUnit.SECONDS);
```
jhaeyaert, line 20: The test looks good, however, using TestScheduler would be better than using await.

jhaeyaert (changes requested):

gravitee-node-container/src/test/java/io/gravitee/node/container/spring/env/WatchablePropertyResolverSubscriberTest.java
lgw-gravitee: sure, I wasn't aware of such tool. I fixed.

jhaeyaert (approved):

PR #345, opened 2024-07-24
fix: NPE using rxPut and rxCompute

jhaeyaert:
**Issue**

https://gravitee.atlassian.net/browse/ARCHI-392

**Description**

This PR fixes possible `NullPointerException` when using reactive `rxPut` and `rxCompute` methods which may return null value that is not allowed with `Single`. 

BREAKING CHANGE: Method signatures have changed from `Single` to `Maybe`. An empty Maybe means 'null' in the blocking world.

---

 - use it directly by updating your project with version: `6.0.0-archi-392-fix-rx-methods-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/6.0.0-archi-392-fix-rx-methods-SNAPSHOT/gravitee-node-6.0.0-archi-392-fix-rx-methods-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
@@ -32,107 +32,317 @@ import java.util.function.Function;
  * @author GraviteeSource Team
  */
 public interface Cache<K, V> {
+    /**
+     * Get the name of this cache.
+     *
+     * @return the name if this cache.
+     */
     String getName();
 
+    /**
+     * Get a collection of all the values in this cache.
+     *
+     * @return a collection of all the values in the cache.
+     */
     Collection<V> values();
 
+    /**
+     * Reactive method to get all the values in this cache as a <code>Flowable</code>.
+     *
+     * @return a <code>Flowable</code> containing all the values in the cache.
+     */
     default Flowable<V> rxValues() {
         return Flowable.fromIterable(this.values()).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Get a set of all the keys in this cache.
+     *
+     * @return a set of all the keys in the cache.
+     */
     Set<K> keys();
 
+    /**
+     * Reactive method to get all the keys in this cache as a <code>Flowable</code>.
+     *
+     * @return a <code>Flowable</code> containing all the keys in the cache.
+     */
     default Flowable<K> rxKeys() {
         return Flowable.fromIterable(this.keys()).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Get set of all the entries in the cache.
+     *
+     * @return all the entries in the cache.
+     */
     Set<Map.Entry<K, V>> entrySet();
 
+    /**
+     * Reactive method to get all the entries in this cache as a <code>Flowable</code>.
+     *
+     * @return a <code>Flowable</code> containing all the entries in the cache.
+     */
     default Flowable<Map.Entry<K, V>> rxEntrySet() {
         return Flowable.fromIterable(this.entrySet()).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Get the current size of this cache.
+     *
+     * @return the size of the cache.
+     */
     int size();
 
+    /**
+     * Reactive method to get the current size of this cache.
+     *
+     * @return a <code>Single</code> containing the size of the cache.
+     */
     default Single<Integer> rxSize() {
         return Single.fromCallable(this::size).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Check if this cache is empty.
+     *
+     * @return <code>true</code> if the cache is empty, <code>false</code> otherwise.
+     */
     boolean isEmpty();
 
+    /**
+     * Reactive method to check if this cache is empty.
+     *
+     * @return a <code>Single<code> containing <code>true</code> if the cache is empty, <code>false</code> otherwise.
+     */
     default Single<Boolean> rxIsEmpty() {
         return Single.fromCallable(this::isEmpty).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Check if the specified key is in this cache.
+     * @param key the key.
+     *
+     * @return <code>true</code> if the key is in the cache, <code>false</code> otherwise.
+     */
     boolean containsKey(final K key);
 
+    /**
+     * Reactive method to check if the specified key is in this cache.
+     * @param key the key.
+     *
+     * @return a <code>Single<code> containing <code>true</code> if the key is in the cache, <code>false</code> otherwise.
+     */
     default Single<Boolean> rxContainsKey(final K key) {
         return Single.fromCallable(() -> this.containsKey(key)).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Return the value for the specified key or <code>null</code> if this cache does not contain this key.
+     * @param key the key used to get the value.
+     *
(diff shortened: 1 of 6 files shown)


gravitee-node-cache/gravitee-node-cache-common/src/main/java/io/gravitee/node/plugin/cache/common/InMemoryCache.java
```diff
@@ -224,12 +224,12 @@ public void putAll(final Map<? extends K, ? extends V> values) {
...
-    public V computeIfAbsent(final K key, final Function<? super K, ? extends V> mappingFunction) {
+    public V computeIfAbsent(final K key, final Function<? super K, ? extends V> remappingFunction) {
         this.internalCache.asMap()
             .computeIfAbsent(
                 key,
                 k -> {
-                    V applied = mappingFunction.apply(k);
+                    V applied = remappingFunction.apply(k);
                     notifyListeners(k, applied, null);
                     return new ExpiringValue<>(applied, 0);
```
guillaumelamirand, line 234: It seems this method always return a expiring value even if the mapping function returns null; As you change the signature of rxComputeIfAbsent to a maybe, a Maybe is not necessary except you test the applied value and return null if null.
The same logic applies to all compute methods.

```suggestion
                    return applied == null ? null : new ExpiringValue<>(applied, 0);
```

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
```diff
@@ -32,107 +32,317 @@
...
+        return Maybe.fromCallable(() -> this.computeIfAbsent(key, mappingFunction)).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * If the specified key is already associated with a value, attempt to compute the value of the specified key using the given mapping function.
+     * If the remapping function returns <code>null</code>, the mapping is removed.
+     * @param key the key.
+     * @param remappingFunction the function that is executed to associate the new value in case a value was present.
+     *
+     * @return the new value if no value was present or the existing one if it was already present.
```
guillaumelamirand, line 244: It should return the new mapping value (including null) in any case.

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
```diff
@@ -32,107 +32,317 @@
...
+     *
+     * @return a <code>Maybe</code> containing the new value if no value was present or the existing one if it was already present.
+     */
+    default Maybe<V> rxComputeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
+        return Maybe.fromCallable(() -> this.computeIfAbsent(key, mappingFunction)).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * If the specified key is already associated with a value, attempt to compute the value of the specified key using the given mapping function.
+     * If the remapping function returns <code>null</code>, the mapping is removed.
```
guillaumelamirand, line 240: It is not the case in the implementation.

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
```diff
@@ -32,107 +32,317 @@
...
+     */
+    V computeIfAbsent(K key, Function<? super K, ? extends V> remappingFunction);
+
+    /**
+     * Reactive method to compute the value of the specified key using the given mapping function if it is not already associated with a value.
+     * If the remapping function returns <code>null</code>, the mapping is removed.
+     * @param key the key.
+     * @param mappingFunction the function that is executed to associate the new value in case no value was present.
+     *
+     * @return a <code>Maybe</code> containing the new value if no value was present or the existing one if it was already present.
```
guillaumelamirand, line 232:
```suggestion
     * @return a <code>Maybe</code> containing the new computed value or empty if the mapped value is null.
```

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
```diff
@@ -32,107 +32,317 @@
...
-
-    default Single<V> rxComputeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
-        return Single.fromCallable(() -> this.computeIfAbsent(key, mappingFunction)).subscribeOn(Schedulers.io());
+    /**
+     * If the specified key is not already associated with a value, attempt to compute the value of the specified key using the given remapping function.
+     * If the remapping function returns <code>null</code>, the mapping is removed.
+     * @param key the key.
+     * @param remappingFunction the function that is executed to associate the new value in case no value was present.
+     *
+     * @return the new value if no value was present or the existing one if it was already present.
```
guillaumelamirand, line 222: It should return the new mapping value (including null) in any case.

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
```diff
@@ -32,107 +32,317 @@
...
+     */
     V computeIfPresent(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction);
 
-    default Single<V> rxComputeIfPresent(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction) {
-        return Single.fromCallable(() -> this.computeIfPresent(key, remappingFunction)).subscribeOn(Schedulers.io());
+    /**
+     * Reactive method to compute the value of the specified key using the given mapping function if it is already associated with a value.
+     * If the remapping function returns <code>null</code>, the mapping is removed.
```
guillaumelamirand, line 250: Looks like the code never removes the value in case of null.

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
```diff
@@ -32,107 +32,317 @@
...
+        return Maybe.fromCallable(() -> this.computeIfPresent(key, remappingFunction)).subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Compute the value of the specified key using the given mapping function.
+     * If the remapping function returns <code>null</code>, the mapping is removed.
+     * @param key the key.
+     * @param remappingFunction the function that is executed to associate the new value.
+     *
+     * @return the new value.
```
guillaumelamirand, line 266: It should return the new mapping value (including null) in any case.

guillaumelamirand (changes requested):

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
jhaeyaert: As seen together, the expected behavior is the following

`@return the computed value if it was absent (<code>null</code> if the mapping function returns <code>null</code>) or the existing one if it was already present.`

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
jhaeyaert: Fixed 👍

gravitee-node-api/src/main/java/io/gravitee/node/api/cache/Cache.java
jhaeyaert: Fixed

guillaumelamirand (approved):

PR #363, opened 2024-09-03
fix: return unhealthy result when a filtered prob is not found

guillaumelamirand:
**Issue**

https://gravitee.atlassian.net/browse/ARCHI-412

**Description**

When a probe is requested on the /health endpoint, if this probe doesn't exist, the endpoint returns 200 SUCCESS instead of an error. This PR attempts to fix this issue.

**Additional context**

### DIFF
--- a/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/healthcheck/NodeHealthCheckManagementEndpoint.java
+++ b/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/healthcheck/NodeHealthCheckManagementEndpoint.java
@@ -20,14 +20,15 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.gravitee.common.http.HttpMethod;
 import io.gravitee.common.http.HttpStatusCode;
 import io.gravitee.common.http.MediaType;
-import io.gravitee.node.api.healthcheck.Probe;
 import io.gravitee.node.api.healthcheck.ProbeEvaluator;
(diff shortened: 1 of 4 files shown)


remibaptistegio (approved):

gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/healthcheck/NodeHealthCheckManagementEndpoint.java
```diff
@@ -57,31 +58,35 @@ public String path() {
...
+                if (probeIds != null) {
+                    probeIds
+                        .stream()
+                        .filter(probeId -> !results.containsKey(probeId))
+                        .forEach(probeId -> results.put(probeId, Result.unhealthy("probe not found")));
+                }
 
                 HttpServerResponse response = ctx.response();
+                boolean healthyProbe = results.values().stream().allMatch(Result::isHealthy);
                 response.setStatusCode(healthyProbe ? HttpStatusCode.OK_200 : HttpStatusCode.INTERNAL_SERVER_ERROR_500);
```
passionne, line 86: In case of unknown probes, shouldn't be a 404 error rather than a 500 ?

guillaumelamirand: The result can contain existing probes healthy or unhealthy in addition to unknown probes, so 404 won't make it for those cases. It is easier to stick with 500.

PR #368, opened 2024-09-19
fix: removed stacktrace from periodic error message

lgw-gravitee:
fixes AM-3909 - In preparation for introducing Degraded Mode, we noticed a flood of the same messages with useless stack traces. By default, this keeps them on one line. To revert this, the log level can be set to DEBUG.

---

 - use it directly by updating your project with version: `6.4.4-fix-AM-3909-too-verbose-syncmanager-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/6.4.4-fix-AM-3909-too-verbose-syncmanager-SNAPSHOT/gravitee-node-6.4.4-fix-AM-3909-too-verbose-syncmanager-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/handler/NodeMonitoringEventHandler.java
+++ b/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/handler/NodeMonitoringEventHandler.java
@@ -80,7 +80,11 @@ public class NodeMonitoringEventHandler extends AbstractService<NodeMonitoringEv
                         .createOrUpdate(convert(message.content()))
                         .ignoreElement()
                         .onErrorResumeNext(throwable -> {
-                            log.error("Unable to process node infos message", throwable);
+                            if (log.isDebugEnabled()) {
+                                log.error("Unable to process node infos message", throwable);
+                            } else {
+                                log.error("Unable to process node infos message, ex={}", throwable.toString());
+                            }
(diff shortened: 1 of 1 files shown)


leleueri (approved):

jhaeyaert (commented):
I noticed you didn't apply the same logic for
`log.error("Unable to process node infos message", throwable);`

Is it on purpose?

Also, I'm wondering if we could have something that logs the full stack at least once and then switches on logging only the message to avoid the situation you mentioned (I know, it's heavier in terms of implementation).

Out of curiosity, do you have an example of stacktrace you encounter in AM?

lgw-gravitee:
Example:
image

lgw-gravitee:
> I noticed you didn't apply the same logic for
> log.error("Unable to process node infos message", throwable);
> 
> Is it on purpose?
No, I changed this as well

> Also, I'm wondering if we could have something that logs the full stack at least once and then switches on logging only the message to avoid the situation you mentioned (I know, it's heavier in terms of implementation).
I wouldn't add such a thing to core lib. It adds more overhead and makes code more complex. Besides, the stack trace isn't meaningful either way.
Maybe I can add Message.destination to log message to make it more clear?

jhaeyaert (approved):

PR #370, opened 2024-10-04
fix(metrics): hold strong reference to probes results in micrometer gauge

wbabyte:
**Issue**

https://github.com/gravitee-io/issues/issues/10070

**Description**

probes results in NaN for prometheus micrometer. It's because we are not holding a strong reference to the state object
[doc](https://docs.micrometer.io/micrometer/reference/concepts/gauges.html#_why_is_my_gauge_reporting_nan_or_disappearing)

**Additional context**

---

 - use it directly by updating your project with version: `6.4.5-APIM-7113-prometheus-metrics-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/6.4.5-APIM-7113-prometheus-metrics-SNAPSHOT/gravitee-node-6.4.5-APIM-7113-prometheus-metrics-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/DefaultProbeEvaluator.java
+++ b/gravitee-node-monitoring/src/main/java/io/gravitee/node/monitoring/DefaultProbeEvaluator.java
@@ -72,4 +72,8 @@ public class DefaultProbeEvaluator implements ProbeEvaluator {
                 }
             });
     }
+
+    public Map<Probe, Result> getCachedResults() {
+        return lastProbeResults;
+    }
 }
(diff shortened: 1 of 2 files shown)


jhaeyaert (approved):

jhaeyaert:
@wbabyte I think it worth applying it on 6.x and perhaps 5.x versions of node (it depends on which versions of node are used by the supported versions of APIM/AM). wdyt?

wbabyte:
> @wbabyte I think it worth applying it on 6.x and perhaps 5.x versions of node (it depends on which versions of node are used by the supported versions of APIM/AM). wdyt?

@jhaeyaert I think I need only 6.0.x for 4.4.x, this PR will release 6.4.5 for 4.5.x and master 
version 5.x of node contains different implementation that already works

PR #375, opened 2024-10-15
feat: allow unsecured tcp server

benoitgravitee:
**Issue**

https://gravitee.atlassian.net/browse/ARCHI-413

**Description**

We can now create a TCP server without SNI enabled. TCP API won't work, but it will allow unsecured usage of TCP server for native events

**Additional context**


gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/server/VertxServerOptions.java
```diff
@@ -282,8 +281,9 @@ protected final void setupTcp(TCPSSLOptions options, KeyCertOptions keyCertOptio
 
             options.setKeyCertOptions(keyCertOptions);
             options.setTrustOptions(trustOptions);
+        } else {
```
jhaeyaert, line 284: Is is mandatory considering ssl is disabled?

jhaeyaert (approved):
I left a question but LGTM 👍

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/server/VertxServerOptions.java
benoitgravitee: not really, when I tested, the collection was not empty so I emptied it.

jhaeyaert: I believe it is fully ignored so I would keep it simple and not empty it.

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/server/tcp/VertxTcpServerOptions.java
```diff
@@ -30,8 +30,6 @@ public NetServerOptions createNetServerOptions(KeyCertOptions vertxKeyCertOption
         if (this.secured && this.sni) {
             options.setSni(true);
             options.setClientAuth(ClientAuth.valueOf(clientAuth));
```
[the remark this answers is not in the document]
benoitgravitee: Good catch

guillaumelamirand (approved):

PR #378, opened 2024-11-08
feat: service secrets common objects and interfaces

benoitgravitee:
**Issue**

APIM-7412

**Description**

* Update of Secret/SecretMap/SecretURL that will trigger a BC in secret-providers
* Draft version of mock secret provider

**Additional context**

---

 - use it directly by updating your project with version: `7.0.0-APIM-7412-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/7.0.0-APIM-7412-SNAPSHOT/gravitee-node-7.0.0-APIM-7412-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/model/SecretURL.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/model/SecretURL.java
@@ -3,7 +3,10 @@ package io.gravitee.node.api.secrets.model;
 import com.google.common.base.Splitter;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
-import java.util.*;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
@@ -16,18 +19,23 @@ import lombok.NoArgsConstructor;
  */
 public record SecretURL(String provider, String path, String key, Multimap<String, String> query) {
     public static final char URL_SEPARATOR = '/';
+    public static final String URI_KEY_SEPARATOR = ":";
     private static final Splitter urlPathSplitter = Splitter.on(URL_SEPARATOR);
     private static final Splitter queryParamSplitter = Splitter.on('&');
     private static final Splitter queryParamKeyValueSplitter = Splitter.on('=');
-    private static final Splitter keyMapParamValueSplitter = Splitter.on(':');
+    private static final Splitter keyMapParamValueSplitter = Splitter.on(URI_KEY_SEPARATOR.charAt(0));
     public static final String SCHEME = "secret://";
+
+    public static SecretURL from(String url) {
+        return from(url, true);
+    }
 
     /**
      * Parse the string into a {@link SecretURL}
      * <p>
      * the format is : <code>secret://&lt;provider&gt;/&lt;path or name&gt;[:&lt;key&gt;][?option=value1&option=value2]</code>
      * </p>
-     * <li>secret://is mandatory</li>
+     * <li><code>secret://</code> is mandatory is includesScheme is true</li>
      * <li>provider is mandatory and should match a secret provider id</li>
      * <li>"path or name" is mandatory, a free string that can contain forward slashes ('/').
      * If an empty string or spaces are found between two forward slashes (eg. <code>//</code> or <code>/ /</code>) parsing will fail.</li>
@@ -36,15 +44,16 @@ public record SecretURL(String provider, String path, String key, Multimap<Strin
      * Pair are always list as can be specified more than once. If no value is parsed, then <code>true</code> is set</li>
      *
      * @param url the string to parse
+     * @param includesScheme to indicate if URL should contain secret://
      * @return SecretURL object
      * @throws IllegalArgumentException when failing to parse
      */
-    public static SecretURL from(String url) {
+    public static SecretURL from(String url, boolean includesScheme) {
         url = Objects.requireNonNull(url).trim();
-        if (!url.startsWith(SCHEME)) {
+        if (includesScheme && !url.startsWith(SCHEME)) {
             throwFormatError(url);
         }
-        String schemeLess = url.substring(SCHEME.length());
+        String schemeLess = includesScheme ? url.substring(SCHEME.length()) : url.substring(1);
         int firstSlash = schemeLess.indexOf('/');
         if (firstSlash < 0 || firstSlash == schemeLess.length() - 1) {
             throwFormatError(url);
@@ -172,6 +181,5 @@ public record SecretURL(String provider, String path, String key, Multimap<Strin
 
         public static final String WATCH = "watch";
         public static final String KEYMAP = "keymap";
-        public static final String NAMESPACE = "namespace";
     }
 }
--- a/gravitee-node-secrets/gravitee-secret-provider-mock/src/main/java/io/gravitee/node/secrets/plugin/mock/MockSecretProvider.java
+++ b/gravitee-node-secrets/gravitee-secret-provider-mock/src/main/java/io/gravitee/node/secrets/plugin/mock/MockSecretProvider.java
@@ -1,0 +1,119 @@
+package io.gravitee.node.secrets.plugin.mock;
+
+import static io.gravitee.plugin.core.internal.AbstractPluginEventListener.SECRET_PROVIDER;
+
+import io.gravitee.node.api.secrets.SecretProvider;
+import io.gravitee.node.api.secrets.model.SecretEvent;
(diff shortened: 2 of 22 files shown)


gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/StringCheck.java
```diff
@@ -0,0 +1,20 @@
+package io.gravitee.node.api.secrets.runtime;
+
+/**
+ * @author Benoit BORDIGONI (benoit.bordigoni at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public class StringCheck {
+
+    private StringCheck() {
```
guillaumelamirand, line 9: You can lombok annotation NoConstructor with access private to generate it.

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/model/SecretURL.java
```diff
@@ -36,15 +44,16 @@ public record SecretURL(String provider, String path, String key, Multimap<Strin
      * Pair are always list as can be specified more than once. If no value is parsed, then <code>true</code> is set</li>
      *
      * @param url the string to parse
+     * @param includesSchema to indicate if URL should contain secret://
      * @return SecretURL object
      * @throws IllegalArgumentException when failing to parse
      */
-    public static SecretURL from(String url) {
+    public static SecretURL from(String url, boolean includesSchema) {
```
guillaumelamirand, line 51:
```suggestion
    public static SecretURL from(String url, boolean includesScheme) {
```
What do you mean by Schema here? Shouldn't it be scheme?

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/StringCheck.java
```diff
@@ -0,0 +1,20 @@
+package io.gravitee.node.api.secrets.runtime;
+
+/**
+ * @author Benoit BORDIGONI (benoit.bordigoni at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public class StringCheck {
```
guillaumelamirand, line 7: I think you could replace tat by either spring lib or guava library. WDYT?

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/discovery/Definition.java
```diff
@@ -0,0 +1,21 @@
+package io.gravitee.node.api.secrets.runtime.discovery;
```
guillaumelamirand, line 1: I am honestly wondering if this shouldn't be somewhere else as it is implemented only by a APIM plugin not a node plugin.

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/discovery/DefinitionSecretRefsFinder.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    boolean canHandle(Object definition);
+
+    /**
+     * Return a description of the definition to be used when checking ACLs
+     * @param definition the definition object to analyse
+     * @param metadata additional properties related to the definition
+     * @return a descriptor of the definition
+     */
+    DefinitionDescriptor getDefinitionLocation(T definition, Map<String, String> metadata);
```
guillaumelamirand, line 25: the method and the return object are not aligned: location vs descriptor, which makes the code more difficult to understand.

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/spec/Spec.java
```diff
@@ -0,0 +1,205 @@
...
+    boolean usesDynamicKey,
+    boolean isGenerated,
+    Resolution resolution,
+    Spec.OnErrorStrategy onErrorStrategy,
+    ACLs acls,
+    String envId
+) {
+    private static final char[] ALLOWED_NON_ALPHA_NUM_CHARS = new char[] { '_', ' ', '.', '-' };
+    static {
+        Arrays.sort(ALLOWED_NON_ALPHA_NUM_CHARS);
```
guillaumelamirand, line 50: why does the order matter?

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/spec/Spec.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    public Spec {
+        boolean ok = isGenerated ^ isNotBlank(id);
+        if (!ok) {
+            throw new IllegalArgumentException("spec is either generated or contains an id");
+        }
+        ok = usesDynamicKey ^ isNotBlank(key);
+        if (!ok) {
+            throw new IllegalArgumentException("spec either uses dynamic key of specifies one");
```
guillaumelamirand, line 60:
```suggestion
            throw new IllegalArgumentException("spec either uses dynamic key or specifies one");
```

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/spec/Spec.java
```diff
@@ -0,0 +1,205 @@
...
+            assertName(name);
+        }
+    }
+
+    /**
+     * Return uri and key concatenated as specified in {@link Spec#formatUriAndKey(String, String)}
+     * @return a string concat of uri and key
+     */
+    public String uriAndKey() {
+        return formatUriAndKey(uri, key);
```
guillaumelamirand, line 84: I don't think you need this redirection, it could be easier to understand if the code was mode here directly.

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/spec/Spec.java
```diff
@@ -0,0 +1,205 @@
...
+        if (key == null) {
+            return uri;
+        }
+        return uri.concat(SecretURL.URI_KEY_SEPARATOR).concat(key);
+    }
+
+    /*
+     * char by analysis to avoid using regexp on a user input
+     */
+    private void assertName(@Nonnull String name) {
```
guillaumelamirand, line 157: I feel like is too much complicate and could be simplify using a simple reg ex. Why do you want to avoid using one?

gravitee-node-secrets/gravitee-secret-provider-mock/src/main/java/io/gravitee/node/secrets/plugin/mock/MockSecretProvider.java
```diff
@@ -0,0 +1,121 @@
...
+                if (Objects.equals(event.error(), "null")) {
+                    return Single.just(new SecretEvent(event.type(), SecretMap.of(event.data())));
+                } else {
+                    return Single.error(new MockSecretProviderException(event.error()));
+                }
+            });
+    }
+
+    @Override
+    public SecretMount fromURL(SecretURL url) {
```
guillaumelamirand, line 118: Should you check the url ?

guillaumelamirand (changes requested):

guillaumelamirand:
Few comments but I am wondering if the code shouldn't put somewhere else as it is not a node feature.

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/spec/Spec.java
benoitgravitee: I'll use io.gravitee.common.utils.IdGenerator to compare it is valid

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/discovery/Definition.java
benoitgravitee: As discussed, this can be used by AM and APIM and the plugin, hence it is here...

gravitee-node-secrets/gravitee-secret-provider-mock/src/main/java/io/gravitee/node/secrets/plugin/mock/MockSecretProvider.java
benoitgravitee: It is usually parse from a string, so no I don't, but I will

gravitee-node-api/src/main/java/io/gravitee/node/api/secrets/runtime/spec/Spec.java
benoitgravitee: The static method is used in plugin and here.

guillaumelamirand (approved):

PR #398, opened 2025-01-10
fix(license-model.yml): add pack "native-kafka" to tier "universe"

passionne:
**Description**

This add the pack "native-kafka" to tier "universe".

⚠ before merge check if this is wanted or not.


phiz71:
Are we sure we want this ?

passionne:
In fact, the current behavior is expected (https://github.com/gravitee-io/gravitee-node/pull/387). Thank you @phiz71 !

PR #434, opened 2025-07-04
fix: added traceparent getter in opentelemetry span

aqibmohammadkhan:
**Issue**

https://github.com/gravitee-io/issues/issues/APIM-9423
**Description**

A small description of what you did in that PR.

**Additional context**

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/Span.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/Span.java
@@ -26,6 +26,11 @@ public interface Span {
     boolean isRoot();
 
     /**
+     * @return traceparent value.
+     */
+    String getTraceparent();
+
+    /**
      * Append custom attribute to the span while it is already created.
      *
      * @return current {@link Span}
--- a/gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/tracer/noop/NoOpSpan.java
+++ b/gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/tracer/noop/NoOpSpan.java
@@ -42,6 +42,11 @@ public class NoOpSpan implements Span {
     }
 
     @Override
+    public String getTraceparent() {
+        return "";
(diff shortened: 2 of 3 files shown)


jhaeyaert (commented):
I understand that it is a backport from a fix on master, but I would like to understand the context where this method is invoked. Can you provide details on how this is supposed to be used and where/how to plan to use it? 🙏

wbabyte:
@jhaeyaert this is to make it possible to propagate traceparent header in the HTTP Callout policy. This header is not available on policy chain, so the only way is to add it to the properties

wbabyte:
in V4 the idea is to add it here https://github.com/gravitee-io/gravitee-api-management/blob/0d061700af22999c53ac5929c3685df8e302464c/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/v4/DefaultApiReactor.java#L323 as an attribute 
like `ctx.setAttribute("traceparent", executionPhaseSpan.getTraceparent());` but if you have better idea happy to improve it

aqibmohammadkhan:
for V3 - same logic (need to create a request span here and extract traceparent and pass in ctx.setAttribute("traceparent", executionPhaseSpan.getTraceparent())

https://github.com/gravitee-io/gravitee-api-management/blob/0d061700af22999c53ac5929c3685df8e302464c/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/handlers/api/ApiReactorHandler.java#L143

for v2 Apis with v4 emulation - need to add similar to startPhaseTracing here
https://github.com/gravitee-io/gravitee-api-management/blob/9e621fbf8e9e8b243e46bcb42614b53682d2a042/gravitee-apim-gateway/gravitee-apim-gateway-handlers/gravitee-apim-gateway-handlers-api/src/main/java/io/gravitee/gateway/reactive/handlers/api/SyncApiReactor.java#L238

and do the same.

PR #547, opened 2026-04-14
feat(redis): add VertxRedisClientFactory for shared Redis clients

wbabyte:
## Summary

Adds `VertxRedisClientFactory` to `gravitee-node-vertx`. The factory does one thing: **deduplicate Vert.x Redis clients by connection tuple with reference counting**. Multiple consumers pointing at the same Redis share one underlying client and pool.

Part of APIM-13603.

## API

```java
Redis client = factory.acquire(options);   // returns shared client; ref-count++
// ... use client ...
factory.release(options);                  // ref-count--; close + remove at zero
```

## Dedup key (minimal, extensible)

- `host`, `port`
- `useSsl`
- `username`
- sentinel `masterId` + sorted `nodes`

**Excluded from key:** password, SSL material, pool config, timeouts. First-acquire wins for these — consumers sharing a connection tuple are assumed to share creds and SSL config. Holds for the current consumer set (cache-redis resource, rate-limit, distributed-sync, node-cache). Key is non-breaking to extend if that assumption changes for a future consumer.

**Security:** key is kept internal and never logged. Log statements include only host/port/useSsl/sentinel topology.

## Out of scope

- **Named configurations** (`redis.configurations.*` in gravitee.yml) — deliberately NOT implemented. Blocked by: console UI can't read gateway config, cloud customers can't edit gateway YAML, and the config reuse problem is a product/UX question better solved at the APIM layer.
- **Global pool defaults** — callers read globals from their own scope (e.g. `ratelimit.redis.*`) and plug them into `RedisClientOptions` before calling `acquire()`. The existing cache-redis plugin already does this.

## SSL

Factory applies SSL to `NetClientOptions` using the node SSL classes (same pattern as `VertxHttpClientFactory`). Base64 decoding of JKS/PKCS12 content lives in the node SSL classes' `trustOptions()` / `keyCertOptions()` methods — one implementation, reused by HTTP, TCP, WebSocket, Redis.

Depends on gravitee-io/gravitee-plugin-common-configurations#14.

## Test plan

- [x] **22 unit tests**: options build (standalone/sentinel/SSL/pool/timeouts); dedup (host/port/ssl/username/sentinel nodes + sentinel-order insensitivity); first-wins for password/pool; ref counting (close-on-zero, create-after-full-release, no-op release of unknown, null-safety); concurrency (16 threads acquiring same tuple → single client)
- [x] All 161 `gravitee-node-vertx` tests pass

---

 - use it directly by updating your project with version: `9.0.0-feat-redis-client-factory-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.0.0-feat-redis-client-factory-SNAPSHOT/gravitee-node-9.0.0-feat-redis-client-factory-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
+++ b/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
@@ -1,0 +1,255 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.vertx.client.redis;
+
+import com.google.common.annotations.VisibleForTesting;
+import io.gravitee.node.vertx.client.ssl.KeyStore;
+import io.gravitee.node.vertx.client.ssl.TrustStore;
+import io.gravitee.plugin.configurations.redis.HostAndPort;
+import io.gravitee.plugin.configurations.redis.RedisClientOptions;
+import io.gravitee.plugin.configurations.redis.RedisSentinelOptions;
+import io.gravitee.plugin.mappers.RedisClientOptionsMapper;
+import io.gravitee.plugin.mappers.SslOptionsMapper;
+import io.vertx.core.Vertx;
+import io.vertx.core.net.NetClientOptions;
+import io.vertx.redis.client.Redis;
+import io.vertx.redis.client.RedisOptions;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Factory for Vert.x Redis clients. Two usage modes, both going through the same
+ * {@link RedisClientOptions} + {@code RedisClientOptionsMapper} + node SSL plumbing,
+ * so consumers don't duplicate the SSL / sentinel / base64 / pool-mapping code.
+ *
+ * <p><b>Shared client (opt-in, ref-counted):</b> {@link #acquire(RedisClientOptions)} /
+ * {@link #release(RedisClientOptions)}. The factory dedups by connection-identifying
+ * fields so multiple consumers pointing at the same endpoint share one underlying
+ * Vert.x client and pool. When the last consumer releases, the client is closed.
+ * Use this for per-API consumers that would otherwise fan out (e.g. the cache-redis
+ * resource: 5000 APIs → 1 shared connection).
+ *
+ * <p><b>Dedicated client (default):</b> {@link #createClient(RedisClientOptions)}
+ * returns a fresh Vert.x Redis client. No dedup, no ref counting, not tracked by the
+ * factory — the caller owns the lifecycle and must close it. Use this for
+ * gateway-singleton consumers (rate-limit, distributed-sync, node-cache-plugin-redis)
+ * that already have one connection per gateway by Spring-bean lifecycle and don't
+ * need or want cross-consumer sharing.
+ *
+ * <p><b>Dedup key:</b>
+ * <ul>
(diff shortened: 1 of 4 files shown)


gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Build the dedup key: host, port, useSsl, username, password, and sentinel
+     * topology. The key is kept internal to this class and never logged.
+     */
+    private static String dedupKey(RedisClientOptions options) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(options.getHost()).append(':').append(options.getPort());
+        sb.append("|ssl=").append(options.isUseSsl());
+        sb.append("|user=").append(nullSafe(options.getUsername()));
+        sb.append("|pw=").append(nullSafe(options.getPassword()));
```
paulatulis, line 212: The password is part of the dedup key String that lives in the ConcurrentHashMap key set for the entire lifetime of the factory (and thus the gateway process). String in Java is interned and GC-unfriendly for secrets — it cannot be zeroed. In a multi-tenant gateway scenario, this is a credential-retention risk.

Recommendation: Hash the password for the key, e.g.:
```
sb.append("|pw=").append(options.getPassword() == null ? "" : Integer.toHexString(options.getPassword().hashCode()));
```

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * <p><b>Invariant:</b> callers using the same connection tuple are assumed to use the
+ * same SSL material and pool config. First acquire wins for those non-keyed fields.
+ * The key can be extended later in a non-breaking way if that assumption changes for
+ * a future consumer.
+ *
+ * <p><b>Security note:</b> the dedup key is kept internal to this class and never logged.
+ * Log statements include only non-sensitive identifying fields (host, port, useSsl).
+ *
+ * <p>Injectable as a Spring bean; {@code Vertx} is supplied via constructor.
```
paulatulis: There is no `@Component` annotation. It must be explicitly registered somewhere (a `@Bean` method in a `@Configuration` class) or the Javadoc is misleading.

Action needed: Either add `@Component` to the class (if component scanning covers this package) or add a note in the Javadoc pointing to where it's registered. If callers are expected to instantiate it themselves, remove the "Spring bean" language.

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
```diff
@@ -0,0 +1,251 @@
...
+     */
+    public Redis createClient(RedisClientOptions options) {
+        if (options == null) {
+            throw new IllegalArgumentException("RedisClientOptions must not be null");
+        }
+        log.info("Created dedicated Redis client for {}", summary(options));
+        return Redis.createClient(vertx, buildRedisOptions(options));
+    }
+
+    RedisOptions buildRedisOptions(RedisClientOptions options) {
```
paulatulis, line 164: Package-private (default visibility) was almost certainly chosen to allow test access without making it `public`. This is a pragmatic tradeoff, but it means any class in the same package can call it — and `RedisOptions` are mutable Vert.x objects. Consider:

Adding a `@VisibleForTesting` annotation (Guava / project equivalent) to document intent.

Or making `buildRedisOptions` private and extracting a test-specific assertion helper.

Same for `sharedClientCount()`

paulatulis (commented):
Passing along some comments from Claude ;)

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
wbabyte: key is never logged, entry is removed from ConcurrentHashMap on last release so the String becomes GC-eligible and callers hold the password String in their own options object for the same lifetime anyway. Integer.toHexString do not protect it in any way

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
wbabyte: Fair call will update the Javadoc 👍

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
wbabyte: added @VisibleForTesting

paulatulis (approved):
LGTM!

phiz71 (approved):

PR #548, opened 2026-04-14
feat: add span-attribute redaction support

vikrantgravitee:
**Issue**

https://gravitee.atlassian.net/browse/APIM-13568

**Description**

Introduce RedactSpanExporter to wrap any SpanExporter and redact span/event attributes before export
Add redaction APIs: RedactionConfig, RedactionRule, MaskingStrategy supporting:
Full masking
Partial masking (prefix/suffix retention)
Pattern-based rules
Enable phase-scoped redaction (REQUEST, RESPONSE, ERROR, ALL) using gravitee.phase
ERROR rules also apply to spans with StatusCode.ERROR
Redact resource attributes at tracer creation via RedactSpanExporter#redactResource (e.g., hostname, IP)
Extend TracerFactory with a 7-arg createTracer overload accepting RedactionConfig (implemented in OpenTelemetryFactory)
Implement single-pass lazy-backfill algorithm with zero allocation when no rules match

**Additional context**

---

 - use it directly by updating your project with version: `9.0.0-feat-redact-span-attribute-otel-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.0.0-feat-redact-span-attribute-otel-SNAPSHOT/gravitee-node-9.0.0-feat-redact-span-attribute-otel-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/redaction/RedactionRule.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/redaction/RedactionRule.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.api.opentelemetry.redaction;
+
+import java.util.Objects;
+
+public record RedactionRule(String attributeNamePattern, MaskingStrategy maskingStrategy, String valuePattern) {
+    public static final String DEFAULT_REPLACEMENT = "[REDACTED]";
+
+    public RedactionRule {
+        Objects.requireNonNull(attributeNamePattern, "attributeNamePattern must not be null");
+        maskingStrategy = (maskingStrategy == null) ? MaskingStrategy.DEFAULT : maskingStrategy;
+    }
+
+    public RedactionRule(String attributeNamePattern) {
+        this(attributeNamePattern, MaskingStrategy.DEFAULT, null);
+    }
+
+    public RedactionRule(String attributeNamePattern, String replacement) {
+        this(attributeNamePattern, MaskingStrategy.fullMask(replacement), null);
+    }
+
+    public RedactionRule(String attributeNamePattern, MaskingStrategy maskingStrategy) {
+        this(attributeNamePattern, maskingStrategy, null);
+    }
+}
--- a/gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/OpenTelemetryFactory.java
+++ b/gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/OpenTelemetryFactory.java
@@ -18,8 +18,10 @@ package io.gravitee.node.opentelemetry;
 import io.gravitee.node.api.opentelemetry.InstrumenterTracerFactory;
 import io.gravitee.node.api.opentelemetry.Tracer;
 import io.gravitee.node.api.opentelemetry.TracerFactory;
+import io.gravitee.node.api.opentelemetry.redaction.RedactionConfig;
 import io.gravitee.node.opentelemetry.configuration.OpenTelemetryConfiguration;
 import io.gravitee.node.opentelemetry.exporter.SpanExporterFactory;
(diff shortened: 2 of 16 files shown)


gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/redaction/Phase.java
```diff
@@ -0,0 +1,39 @@
...
+ *       {@code gravitee.phase} attribute is explicitly set to the matching value by the
+ *       producing instrumentation. Spans without that attribute (or with an unrecognised value)
+ *       do <em>not</em> match these phases.</li>
+ *   <li>{@link #ERROR} — evaluated when the span ended with
+ *       {@link io.opentelemetry.api.trace.StatusCode#ERROR}, regardless of which execution
+ *       phase the span belongs to. An error span in the REQUEST phase matches both
+ *       REQUEST-scoped rules and ERROR-scoped rules simultaneously.</li>
+ * </ul>
+ */
+public enum Phase {
```
remibaptistegio: I'm not sure we should have an Enum here to map phases. Phase are a concept that is known by APIM only and REQUEST/RESPONSE will not match in other scenarios no ? Maybe we should keep it as simple as possible in node by only managing a key/value in the rule to determine if it should be applied or not.
And when you configure it from APIM, in the UI, you could have a Helper that only list the "known" phases and the right attribute to compare to, this way we could implement the same in AM for instance.

gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/redaction/RedactionRule.java
```diff
@@ -0,0 +1,60 @@
...
+ * <p>The {@code valuePattern} field (when non-null) is a Java regular expression evaluated
+ * via {@link java.util.regex.Matcher#find()} — i.e. a <em>substring</em> match, not a
+ * full-string match. Use {@code ^…$} anchors when you need an exact match.
+ *
+ * <p><strong>Null-asymmetry:</strong> when {@code valuePattern} is {@code null} (no pattern),
+ * the rule redacts <em>every</em> attribute whose key matches {@code attributeNamePattern},
+ * regardless of its value. When {@code valuePattern} is non-null, an attribute with a
+ * {@code null} value is <em>never</em> redacted (a null value cannot match any regex).
+ */
+public record RedactionRule(String attributeNamePattern, MaskingStrategy maskingStrategy, String valuePattern, Phase phase) {
```
remibaptistegio: Just to be sure, so if I want to redact all values that looks like a Mastercard number, I should create a rule with attributeNamePatter = "*" and valuePattern = "(5[1-5][0-9]{14})" ? I need an attributeNamePattern, I cannot just define a valuePattern, right ?

gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/OpenTelemetryFactory.java
```diff
@@ -66,7 +69,15 @@ public Tracer createTracer(
...
-        return createTracer(serviceInstanceId, serviceName, serviceNamespace, serviceVersion, instrumenterTracerFactories, null);
+        return createTracer(
+            serviceInstanceId,
+            serviceName,
+            serviceNamespace,
+            serviceVersion,
+            instrumenterTracerFactories,
+            null,
+            RedactionConfig.EMPTY
+        );
```
remibaptistegio, line 80: Why don't we read the redaction configuration in the OpenTelemetryConfiguration, so we can benefit from the redaction feature without changing something in APIM for instance ?
Let's say I have my gravitee.yaml like this:
```
services:
  opentelemetry:
    enabled: false
    verbose: false
    exporter:
      endpoint: http://localhost:4317
      protocol: grpc
    redactionRules:
      #rule mastercard
      - attributeNamePatter: *
        maskingStrategy:
          type: FULL
          replacement: [NOT_DISPLAYED]
        valuePattern: (5[1-5][0-9]{14})
```

gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/redaction/RedactionRule.java
vikrantgravitee: For now - yes.
attributeNamePattern is required.

- attributeNamePattern: "**"
  valuePattern: "(5[1-5][0-9]{14})"

gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/OpenTelemetryFactory.java
vikrantgravitee: Great point. This is absolutely the right approach and we should implement it.

gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/redaction/Phase.java
vikrantgravitee: I agree to this point and missed the AM. I will remove this Phase dependency from gravitee-node.

remibaptistegio (approved):

PR #549, opened 2026-04-16
refactor(node-cache-redis): migrate to VertxRedisClientFactory

wbabyte:
## Summary
- Migrates `gravitee-node-cache-plugin-redis` to use `VertxRedisClientFactory.createClient(RedisClientOptions)` instead of its own `RedisOptionsFactory` + `Redis.createClient()`.
- Deletes 4 redundant config POJOs (`RedisOptionsFactory`, `RedisConfiguration`, `HostAndPort`, `SentinelConfiguration`) — replaced by `RedisClientOptions` from `gravitee-plugin-common-configurations`.
- `RedisConfigurationProvider.from()` rewritten to return `RedisClientOptions` directly, reading the same `cache.redis.*` YAML properties.
- SSL/sentinel/pool option-building now routes through the factory's shared plumbing (no more duplicated code).

### Bundled behaviour change
The legacy `RedisConfigurationProvider` called `SentinelConfiguration#setMaster(...)` twice — the second call silently clobbered the master id with the password value, so sentinel + password configs were broken. The rewrite calls `masterId(...)` and `password(...)` on the new builder, so both fields land where they should. Covered by `RedisConfigurationProviderTest#Sentinel#should_map_master_and_password_to_distinct_fields`.

Stacked on PR #547 (`feat/redis-client-factory`).

## Test plan
- [x] `RedisCacheTest` (5/5) — embedded Redis, validates get/put/evict/clear/TTL lifecycle
- [x] `RedisCacheConnectionFailureTest` (5/5) — validates error handling on unreachable host
- [x] `RedisConfigurationProviderTest` (24/24) — YAML → `RedisClientOptions` mapping: defaults, standalone, pool tuning, sentinel (master/password distinct fields, default `mymaster`, indexed node loop), SSL (skip when disabled, `openssl` propagation, hostname-verifier NONE→false / HTTPS→true, JKS/PKCS12/PEM keystore + truststore matrix, unknown type → none), prefix isolation
- [ ] Integration: enable `cache.type: redis` in `gravitee.yml`, start gateway, confirm round-trip

---

 - use it directly by updating your project with version: `9.0.0-wba-node-cache-plugin-redis-migration-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.0.0-wba-node-cache-plugin-redis-migration-SNAPSHOT/gravitee-node-9.0.0-wba-node-cache-plugin-redis-migration-SNAPSHOT.zip)


gravitee-node-cache/gravitee-node-cache-plugin-redis/src/main/java/io/gravitee/node/plugin/cache/redis/configuration/RedisConfigurationProvider.java
```diff
@@ -53,56 +59,51 @@ private RedisConfigurationProvider() {
...
+            final RedisSentinelOptions.RedisSentinelOptionsBuilder sentinelBuilder = RedisSentinelOptions.builder();
+            sentinelBuilder.masterId(environment.getProperty(propertiesPrefix + SENTINEL_PREFIX + ".master", String.class, "mymaster"));
+            sentinelBuilder.password(environment.getProperty(propertiesPrefix + SENTINEL_PREFIX + "." + PASSWORD_KEY, String.class));
+            sentinelBuilder.nodes(getSentinelNodes(environment, propertiesPrefix));
+            builder.sentinel(sentinelBuilder.build());
         }
-        config.setSentinelConfiguration(sentinelConfiguration);
 
         if (useSsl) {
-            final var useOpenSSL = environment.getProperty(propertiesPrefix + ".openssl", Boolean.class, false);
```
paulatulis, line 82: Do we need to call out this being dropped as a breaking change in release notes?

gravitee-node-cache/gravitee-node-cache-plugin-redis/src/main/java/io/gravitee/node/plugin/cache/redis/spring/RedisCacheConfiguration.java
```diff
@@ -30,7 +32,12 @@
 public class RedisCacheConfiguration {
```
paulatulis, line 32: RedisCacheConfiguration declares the factory bean in the cache-redis plugin module:
```
  @Bean
  public VertxRedisClientFactory vertxRedisClientFactory(Vertx vertx) {
      return new VertxRedisClientFactory(vertx);
  }
```
But #547 added `VertxRedisClientFactory` to gravitee-node-vertx precisely so multiple consumers (cache-redis, rate-limit, distributed-sync, node-cache) share one factory instance and therefore share clients. If every plugin instantiates its own factory bean, the dedup/ref-counting in #547 only dedups within a single plugin — defeating the point. The bean should live in a shared module (e.g. gravitee-node-vertx's Spring config) so all plugins get the same singleton. 

Can you confirm this isn't intended to be moved later in the stac? If it is, fine; if not, this PR is silently breaking the cross-plugin sharing.

gravitee-node-cache/gravitee-node-cache-plugin-redis/src/main/java/io/gravitee/node/plugin/cache/redis/RedisCacheManager.java
```diff
@@ -89,7 +87,7 @@ protected void doStop() {
 
     private synchronized RedisAPI getOrCreateRedisAPI() {
```
paulatulis, line 89: `doStop()` doesn't release / null out the API

Not visible in the diff but in the file: getOrCreateRedisAPI() caches redisAPI in a field, and doStop() (existing code) presumably closes it. After the migration:
  - If the factory uses acquire/release, doStop needs factory.release(options) — otherwise the client leaks and the ref-count never hits zero.
  - The cached redisAPI field isn't nulled either way, so a stop-then-start cycle on the same RedisCacheManager would return a closed/stale RedisAPI. May have been broken pre-PR — worth a glance.

paulatulis (commented):
Does the cache-redis migration participate in client sharing with rate-limit / distributed-sync / node-cache, or does it hold its own dedicated client?

Do you want to add tests for sentinel or SSL paths other than standalone happy path?

gravitee-node-cache/gravitee-node-cache-plugin-redis/src/main/java/io/gravitee/node/plugin/cache/redis/configuration/RedisConfigurationProvider.java
wbabyte: Three things were silently dropped in the original push; two are now resolved:

- `cache.redis.openssl` — **restored in f0d8a931**. Reads the YAML key again and flows through to `SslOptions.openSsl`. No breaking change.
- `hostnameVerificationAlgorithm: LDAPS` (and any non-`NONE`/`HTTPS` value) — collapsed to a boolean and coerced to `"HTTPS"` downstream by `SslOptions.effectiveHostnameVerificationAlgorithm()`. Schema-forced by the new `plugin-common-configurations` `SslOptions` (boolean-only field). Minor — `LDAPS` is meaningless for a Redis endpoint anyway. Now documented in the inline comment. No release note needed.
- Sentinel master/password bug fix — the old code overwrote `master` with the password value via two `setMaster()` calls. The new builder maps them to distinct fields. This does change runtime behaviour for anyone who had sentinel + password configured (their setup was broken and will start working). Documented in the PR description and covered by `RedisConfigurationProviderTest.Sentinel#should_map_master_and_password_to_distinct_fields`. I'll add a dedicated `fix:` commit so it lands as a changelog entry.

gravitee-node-cache/gravitee-node-cache-plugin-redis/src/main/java/io/gravitee/node/plugin/cache/redis/spring/RedisCacheConfiguration.java
wbabyte: Valid concern, and explicitly intended to be moved later in the stack.

- This PR's path is `createClient` (dedicated, not ref-counted) — `node-cache-plugin-redis` is a gateway singleton (one Redis connection per gateway via Spring bean lifecycle). Whether the factory instance is shared with other plugins is irrelevant for the dedicated path; the result is identical.
- Cross-plugin sharing only kicks in when consumers call `acquire`/`release` — none of `node-cache-plugin-redis`, `rate-limit`, `distributed-sync` do. The dedup target is the per-API `gravitee-resource-cache-redis` plugin (separate repo, ~5000 APIs → 1 client). When it migrates onto the factory it'll use `acquire`/`release`.
- Long-term home for the bean: `gravitee-node-vertx` autoconfig (`io.gravitee.node.vertx.spring`), so all plugins resolve the same singleton via injection. Tracking as a follow-up under epic APIM-13553. Not blocking for this PR because it ships a `createClient`-only consumer.

So: not silently breaking — the sharing path doesn't exist yet on this branch. Will move the bean before any `acquire`-using consumer lands.

gravitee-node-cache/gravitee-node-cache-plugin-redis/src/main/java/io/gravitee/node/plugin/cache/redis/RedisCacheManager.java
wbabyte: Two parts:

- **`factory.release(options)` not needed.** `RedisCacheManager` uses `createClient` (dedicated), not `acquire`. The factory doesn't track the client, so there's nothing to release. `redisAPI.close()` in `doStop()` closes the underlying Vert.x client directly — correct lifecycle for the dedicated path.
- **`redisAPI` not nulled after `close()`** — real pre-existing bug, agreed. Stop-then-start returns a closed `RedisAPI`. Will fix in this PR with a follow-up commit so a restart cycle gets a fresh client.

wbabyte:
**Sharing**: `node-cache-plugin-redis` does **not** participate in cross-consumer sharing — holds its own dedicated client via `createClient`. Same intent for `rate-limit` and `distributed-sync` (each is already a gateway singleton). The only consumer designed to `acquire`/`release` is the per-API `gravitee-resource-cache-redis` (separate repo, Phase 1 already ships intra-plugin reference counting; Phase 2 migrates it onto this factory).

**Sentinel / SSL tests**: added in b61f6a59. `RedisConfigurationProviderTest` has 24 cases: sentinel master/password distinct fields (regression for the silent fix), default master id, indexed node loop with gap, disabled when no nodes; SSL skip when `ssl=false`, `openssl` propagation, hostname-verifier NONE/HTTPS/LDAPS mapping (parameterized), JKS/PKCS12/PEM keystore + truststore matrix, unknown type → none, prefix isolation. Standalone happy path still in `RedisCacheTest` against embedded Redis.

paulatulis (approved):
Good to go from my end.

PR #554, opened 2026-04-29
feat: implement a tracing query service based on Tempo

remibaptistegio:
**Issue**

https://github.com/gravitee-io/issues/issues/XXXXX

**Description**

Add a generic tracing query service to fetch traces. Right now, only a Tempo implementation is available but it could be enhanced later. 

---

 - use it directly by updating your project with version: `9.0.0-feat-add-tracing-tempo-query-service-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.0.0-feat-add-tracing-tempo-query-service-SNAPSHOT/gravitee-node-9.0.0-feat-add-tracing-tempo-query-service-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/query/model/Trace.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/query/model/Trace.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.api.opentelemetry.query.model;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Aggregated view of a distributed trace returned by a {@link io.gravitee.node.api.opentelemetry.query.TracingQueryService}.
+ * <p>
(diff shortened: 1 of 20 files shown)


gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/query/model/Trace.java
```diff
@@ -0,0 +1,44 @@
...
+    Instant startTime,
+    long durationNanos,
+    String rootService,
+    String rootOperation,
+    Boolean hasError,
+    List<TraceSpan> spans
+) {
+    public Trace {
+        if (spans == null) {
+            spans = List.of();
```
jhaeyaert: Maybe switch all the List/Map.of() to new ArrayList/Hashmap so they are modifiable if necessary.

gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/configuration/TracingQueryConfiguration.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Binds the {@code services.opentelemetry.query.*} namespace and exposes the {@link SslOptions} / {@link VertxHttpProxyOptions} /
+ * {@link VertxHttpClientOptions} instances expected by {@link io.gravitee.node.vertx.client.http.VertxHttpClientFactory}.
+ *
+ * @author GraviteeSource Team
+ */
+@Getter
+@Setter
+public class TracingQueryConfiguration {
```
jhaeyaert, line 33: Maybe let the application create the configuration itself instead of autoconfiguring it at the node level. That way, we can have APIM instantiating the tracing query as it wants, using their own configuration keys (ex: 'repository.otel.xxx')

jhaeyaert (changes requested):

gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/configuration/TracingQueryConfiguration.java
remibaptistegio: Changes made to provide this kind of mechanism. Now the service is not managed in node, a factory is available to create it using a provider for the configuration.

jhaeyaert (approved):

PR #562, opened 2026-05-18
feat/APIM-13569: payload data masking

araffass:
**Issue**

https://gravitee.atlassian.net/browse/APIM-13576
https://gravitee.atlassian.net/browse/APIM-13578
https://gravitee.atlassian.net/browse/APIM-13579

**Description**

Implements OTel payload body masking for verbose tracing (APIM-13569).

  Adds the data model, masking engine, and export pipeline integration needed to redact sensitive fields from request/response bodies before they are exported to the tracing backend.

  - APIM-13576: PayloadMaskingRule/Config models (path, format, phase, strategy)
  - APIM-13578: JSONPath + XPath masking engine with AUTO format detection
  - APIM-13579: Hook into RedactSpanExporter to mask payload span event bodies before export + global rules from gravitee.yml

**Additional context**

---

 - use it directly by updating your project with version: `9.0.0-feat-APIM-13569-payload-masking-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.0.0-feat-APIM-13569-payload-masking-SNAPSHOT/gravitee-node-9.0.0-feat-APIM-13569-payload-masking-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/TracerFactory.java
+++ b/gravitee-node-api/src/main/java/io/gravitee/node/api/opentelemetry/TracerFactory.java
@@ -15,6 +15,7 @@
  */
 package io.gravitee.node.api.opentelemetry;
 
+import io.gravitee.node.api.opentelemetry.redaction.PayloadMaskingConfig;
 import io.gravitee.node.api.opentelemetry.redaction.RedactionConfig;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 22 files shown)


jhaeyaert:
Can't we implement such a masking behavior by extending the tracing mechanism so a Policy can add the masking behavior? This would be better aligned with what we have with the [Data Logging Masking Policy](https://documentation.gravitee.io/apim/create-and-configure-apis/apply-policies/policy-reference/data-logging-masking)?

remibaptistegio:
> Can't we implement such a masking behavior by extending the tracing mechanism so a Policy can add the masking behavior? This would be better aligned with what we have with the [Data Logging Masking Policy](https://documentation.gravitee.io/apim/create-and-configure-apis/apply-policies/policy-reference/data-logging-masking)?

I guess in fact this policy should already be doing the requested feature no ? Since we are using analytics Logs to produce OTel Logs, if we apply this policy and mask some part of the analytics Logs, then the OTel Log produced should be masked as expected.

brasseld (changes requested):
@araffass can you please answer the questions from @remibaptistegio and @jhaeyaert before going on that one.

araffass:
@brasseld Regarding the two questions, if my underestanding is right: 
The Data Logging Masking Policy operates on the analytics Log object, it masks fields before the log is dispatched through ReporterService to reporters (Elasticsearch, etc.). It doesn't intercept OTel span events, which are emitted via span.addEvent("payload", ...) independently, directly into the OTel SDK pipeline.

Looking at the code, both happen from the same capture point (where i've put the new masking change) :
  this.setBody(capturedBody);           // analytics log pipeline, masked by Data Logging Masking Policy
  emitPayloadSpanEvent(ctx, capturedBody);  // span event, needs its own masking = what this PR adds

So the masking policy wouldn't cover span events regardless of whether OTel logs are also produced from the analytics pipeline.

That said, @jhaeyaert 's point is a fair architectural question: should bodies even be in span events, or should we rely purely on the analytics log → OTel log path?

I've based my implementations on what was on Jira but this stays a design decision that should be aligned on before merging so i'll be happy to adjust if needed / confirmed from your side.

remibaptistegio:
> @brasseld Regarding the two questions, if my underestanding is right: The Data Logging Masking Policy operates on the analytics Log object, it masks fields before the log is dispatched through ReporterService to reporters (Elasticsearch, etc.). It doesn't intercept OTel span events, which are emitted via span.addEvent("payload", ...) independently, directly into the OTel SDK pipeline.
> 
> Looking at the code, both happen from the same capture point (where i've put the new masking change) : this.setBody(capturedBody); // analytics log pipeline, masked by Data Logging Masking Policy emitPayloadSpanEvent(ctx, capturedBody); // span event, needs its own masking = what this PR adds
> 
> So the masking policy wouldn't cover span events regardless of whether OTel logs are also produced from the analytics pipeline.
> 
> That said, @jhaeyaert 's point is a fair architectural question: should bodies even be in span events, or should we rely purely on the analytics log → OTel log path?
> 
> I've based my implementations on what was on Jira but this stays a design decision that should be aligned on before merging so i'll be happy to adjust if needed / confirmed from your side.

After a discussion with @araffass, it turns out that the jira ticket was not reflecting the actual work need regarding this topic (we only need to capture the payload using logs and not span events). So we don't need this and can reuse the data masking policy to complete this feature. This PR can be closed.

brasseld:
Thanks @remibaptistegio @araffass

PR #567, opened 2026-06-02
feat(redis): support Redis Cluster in Vert.x client factory [APIM-14146]

wbabyte:
## APIM-14146 — Enable Redis Cluster (2/3: node client factory)

Part of [APIM-14146](https://gravitee.atlassian.net/browse/APIM-14146).

Teaches the Vert.x Redis client factory about cluster topology. The actual `RedisClientType.CLUSTER` wiring lives in `gravitee-plugin-common-configurations` (PR 1/3); this PR adds the node-side awareness.

### Changes
- `VertxRedisClientFactory.dedupKey()` includes the cluster node set (sorted) so distinct cluster configs don't collide in the shared-client registry.
- `summary()` logs cluster node count (credentials never logged).
- SSL is untouched — the single `NetClientOptions` already applies across all endpoints.

### Tests
`VertxRedisClientFactoryTest` — cluster `buildRedisOptions` + dedup-by-cluster-nodes. Full suite green.

### Release chain
Depends on **PR 1/3** (`gravitee-plugin-common-configurations`). ⚠️ Before merge, bump `gravitee-plugin-common-configurations.version` to the version released from PR 1 (currently set to `1.3.0` for local cross-repo testing). Consumed downstream by `gravitee-api-management` (PR 3/3).

[APIM-14146]: https://gravitee.atlassian.net/browse/APIM-14146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

 - use it directly by updating your project with version: `9.0.0-wba-APIM-14146-redis-cluster-client-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.0.0-wba-APIM-14146-redis-cluster-client-SNAPSHOT/gravitee-node-9.0.0-wba-APIM-14146-redis-cluster-client-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
+++ b/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
@@ -20,6 +20,7 @@ import io.gravitee.node.vertx.client.ssl.KeyStore;
 import io.gravitee.node.vertx.client.ssl.TrustStore;
 import io.gravitee.plugin.configurations.redis.HostAndPort;
 import io.gravitee.plugin.configurations.redis.RedisClientOptions;
+import io.gravitee.plugin.configurations.redis.RedisClusterOptions;
 import io.gravitee.plugin.configurations.redis.RedisSentinelOptions;
 import io.gravitee.plugin.mappers.RedisClientOptionsMapper;
 import io.gravitee.plugin.mappers.SslOptionsMapper;
@@ -60,6 +61,7 @@ import lombok.RequiredArgsConstructor;
  *   <li>{@code username}</li>
  *   <li>{@code password} (never logged)</li>
  *   <li>sentinel {@code masterId} and sorted {@code nodes}</li>
+ *   <li>cluster {@code useReplicas} and sorted {@code nodes}</li>
  * </ul>
  *
  * <p><b>Invariant:</b> callers using the same connection tuple are assumed to use the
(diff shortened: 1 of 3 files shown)


gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
```diff
@@ -233,6 +235,16 @@ private static String dedupKey(RedisClientOptions options) {
                 sb.append(',').append(node.getHost()).append(':').append(node.getPort());
             }
         }
+
+        RedisClusterOptions cluster = options.getCluster();
+        if (cluster != null && cluster.getNodes() != null && !cluster.getNodes().isEmpty()) {
+            sb.append("|cluster=").append(cluster.getUseReplicas());
```
paulatulis: `sb.append("|cluster=").append(cluster.getUseReplicas())` appends the raw String. Via the no-args/@AllArgsConstructor path this can be null → literal "null", and "NEVER" vs null (both resolve to NEVER in the mapper) would key differently → missed sharing (not incorrect). Wrapping in the existing nullSafe(...) / normalizing-to-uppercase would tighten it. Minor.

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
paulatulis, line 1: Latent risk, but not a blocker: dedupKey gates the cluster segment on node-presence only, but the mapper's `isClusterMode` also requires `cluster.isEnabled()`. So two configs that differ only by `cluster.enabled` (same nodes) get the same dedup key yet resolve to different clients (CLUSTER vs STANDALONE) — first-acquire wins and the second consumer silently gets the wrong topology. Not triggerable today (APIM always builds with enabled=true and rate-limit uses createClient, not acquire), but it's a correctness gap in the shared acquire path. Suggest mirroring the mapper's predicate here — add && cluster.isEnabled() — and the same on the sentinel branch. Alternatively, a comment noting the "callers never disable a mode while leaving nodes populated" assumption.

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
wbabyte: Done wrapped in `nullSafe(...)`.

gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/redis/VertxRedisClientFactory.java
wbabyte: Done added `&& isEnabled()` to both the cluster and sentinel segments, mirroring the mapper, plus a test for the enabled-vs-disabled split.

paulatulis (approved):

PR #580, opened 2026-06-29
feat(monitoring): add probe for gpu device monitoring

remisultan:
**Description**

 Adds GPU metrics collection to gravitee-node-monitoring, exposed via Prometheus, the /monitor endpoint, and node alerts alongside the existing OS/JVM/Process metrics.                                                               

  - GpuInfo record (gravitee-node-api): per-device utilization, memory, temperature, power, and identity (index/name/uuid/driver).                                                                                                       
  - Pluggable provider SPI (GpuMetricsProvider) with an nvidia-smi-backed implementation that shells out and parses CSV, no native dependency, matching the existing probe style. The SPI leaves room for other vendors (e.g. NVML,     
  Metal) later.                                                                                                                                                                                                                          
  - Producer/consumer pipeline: NodeGpuMonitorService collects the GPU snapshot once per interval and publishes it on the Vert.x event bus; GpuMonitorEventHandler consumes it into a shared GpuSnapshotRegistry and binds the Micrometer
  gauges. The node monitor, /monitor endpoint, and gauges all read the same snapshot, the GPU is probed only once per cycle.                                                                                                            
  - Configurable cadence via services.monitoring.gpu.{enabled,delay,unit}, defaulting to the node-monitoring settings, `services.monitoring.gpu.enabled` defaults to false not to enable it on other node-based compoments (APIM, AM, ...).                                                                                                                   
  - Performance: collection runs off the hot path on a dedicated thread; availability is probed once and memoized, so each cycle does a single lightweight nvidia-smi query. Hosts without a GPU pay nothing.

<!-- Version placeholder -->

---

 - use it directly by updating your project with version: `9.2.0-feat-metrics-gpu-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.2.0-feat-metrics-gpu-SNAPSHOT/gravitee-node-9.2.0-feat-metrics-gpu-SNAPSHOT.zip)
  <!-- Version placeholder end -->

### DIFF
--- a/gravitee-node-monitoring/src/test/java/io/gravitee/node/monitoring/monitor/gpu/GpuInfoMemTest.java
+++ b/gravitee-node-monitoring/src/test/java/io/gravitee/node/monitoring/monitor/gpu/GpuInfoMemTest.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.monitoring.monitor.gpu;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.gravitee.node.api.monitor.GpuInfo;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @author GraviteeSource Team
+ */
+class GpuInfoMemTest {
+
+    @Test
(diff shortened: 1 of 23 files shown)


gravitee-node-monitoring/src/test/java/io/gravitee/node/monitoring/monitor/gpu/GpuInfoMemTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.gravitee.node.api.monitor.GpuInfo;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @author GraviteeSource Team
+ */
+class GpuInfoMemTest {
```
remibaptistegio, line 29: Could you add `@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)` on your test classes.

remibaptistegio (approved):
Just one comment on test classes otherwise LGTM.

gravitee-node-monitoring/src/test/java/io/gravitee/node/monitoring/monitor/gpu/GpuInfoMemTest.java
remisultan: Done !

PR #583, opened 2026-07-01
fix: evaluate certificate expiry

leleueri:
This PR evaluate the certificate expiry brings by keystores/Truststore

the goal is to help diagnose TLS handshake issue without enabling debug traces

related to AM-6934

---

 - use it directly by updating your project with version: `9.1.2-am-6934-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.1.2-am-6934-SNAPSHOT/gravitee-node-9.1.2-am-6934-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-certificates/src/main/java/io/gravitee/node/certificates/CertificateExpiryUtils.java
+++ b/gravitee-node-certificates/src/main/java/io/gravitee/node/certificates/CertificateExpiryUtils.java
@@ -1,0 +1,148 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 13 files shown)


gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/client/ssl/SslCertificateExpiryInspector.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Best-effort inspection of client-side keystore/truststore material to warn about certificates
+ * that are expired, not yet valid, or about to expire.
+ * <p>
+ * Materialization failures are swallowed (logged at debug level) so that an inspection problem can
+ * never prevent a client from being created.
+ *
+ * @author GraviteeSource Team
+ */
+public final class SslCertificateExpiryInspector {
```
remibaptistegio: Why not use the same utility class for both ? You could add this logic inside the CertificateExpiryUtils no ? There is also some existing code in `KeyStoreUtils` that you could reuse I think.

leleueri: I merge the two classes, but I do not see which method should I use in the KeyStoreUtil.
I already use some `initFromXXX` methods

remibaptistegio: Maybe you could use the `KeyStoreUtils.loadPemCertificates` in your inspectPem method no ?

leleueri: PR updated

remibaptistegio (approved):

PR #587, opened 2026-07-15
fix: read documented plural services.metrics.domains config key

callaertanthony:
## Problem

The gateway silently ignores the documented `services.metrics.domains` (plural) YAML list, so a disabled metrics domain such as `NAMED_POOLS` can never be re-enabled — `pool_queue_pending` / `pool_queue_time_seconds` never appear in the Prometheus output. (APIM-14510)

## Root cause

`VertxFactory.loadMetricsDomains()` reads the **singular** indexed key `services.metrics.domain[N]` via manual `Environment.getProperty`, but the documented/used config key is the **plural** `services.metrics.domains`. Because this is a manual indexed lookup (not relaxed-binding `@ConfigurationProperties`), the plural key is never read, and the code always falls back to the hardcoded default-disabled domains list (which includes `NAMED_POOLS` since 8.0.2).

A known workaround was to use the singular `domain` key, which Spring's `YamlPropertySourceLoader` flattens into exactly the path the code expected.

## Fix

Read the documented plural key `services.metrics.domains[N]`. Also keep reading the singular `services.metrics.domain[N]` for **backward compatibility** with deployments that adopted the singular workaround, so upgrading doesn't silently drop their config.

## Tests

`VertxFactoryTest`:
- `should_read_metrics_domains_from_documented_plural_key`
- `should_still_read_metrics_domains_from_legacy_singular_key`

(The two shared builder stubs were made `lenient()` — consistent with the existing lenient stub — since these direct-method tests don't bootstrap Vert.x.)

`mvn -pl gravitee-node-vertx test -Dtest=VertxFactoryTest` → `Tests run: 16, Failures: 0, Errors: 0`.

---

 - use it directly by updating your project with version: `9.2.3-fix-apim-14510-metrics-domains-plural-key-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.2.3-fix-apim-14510-metrics-domains-plural-key-SNAPSHOT/gravitee-node-9.2.3-fix-apim-14510-metrics-domains-plural-key-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/VertxFactory.java
+++ b/gravitee-node-vertx/src/main/java/io/gravitee/node/vertx/VertxFactory.java
@@ -269,13 +269,21 @@ public class VertxFactory implements FactoryBean<Vertx> {
     private Set<MetricsDomain> loadMetricsDomains() {
         final Set<MetricsDomain> metricsDomains = new HashSet<>();
 
-        String value;
-        int counter = 0;
-        while ((value = environment.getProperty("services.metrics.domain[" + (counter++) + "]")) != null) {
-            metricsDomains.add(MetricsDomain.valueOf(value));
-        }
+        // "services.metrics.domains" (plural) is the documented configuration key. The singular
+        // "services.metrics.domain" is also read for backward compatibility, as some deployments
+        // relied on it as a workaround while the plural key was ignored (APIM-14510).
+        readIndexedMetricsDomains("services.metrics.domains", metricsDomains);
+        readIndexedMetricsDomains("services.metrics.domain", metricsDomains);
 
         return metricsDomains;
+    }
+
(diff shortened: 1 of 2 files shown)


callaertanthony:
Code review — looks good, no blockers.

Supporting the documented plural `domains` while keeping the singular `domain` backward-compat is sound, and collecting into a `Set` de-dupes if both are set. A couple of low-severity nits:
- Tests invoke the private `loadMetricsDomains` via `ReflectionTestUtils` — couples tests to the implementation; testing through the public factory path would be more robust (though heavier to set up).
- Optionally `.trim().toUpperCase()` before `MetricsDomain.valueOf(value)` to tolerate whitespace/case in config.
- Pre-existing: the sequential-index `while` loop stops at the first gap (e.g. `[0]`, `[2]` would drop `[2]`).

remibaptistegio (approved):

PR #601, opened 2026-08-13
feat(opentelemetry): accept base64 keystore and truststore content

josephlomasney:
**Issue**

https://gravitee.atlassian.net/browse/CJ-4304

**Description**

Cockpit is adding an OpenTelemetry reporter form where customers can upload client certificates for mTLS to their own collector. Cockpit already pushes reporter config down to the gateways, so the certificate wants to travel the same way, but the OTLP exporter only reads keystores and truststores from a file path.

This adds `services.opentelemetry.exporter.ssl.{keystore,truststore}.content`, taking a base64-encoded jks or pkcs12 store. PEM is untouched and stays path only. Traces only, since the log record exporter uses the OTel SDK's own sender rather than this one. Same idea as gravitee-io/gravitee-reporter-tcp#131 for the TCP reporter.

A store whose type is set but has no usable path or content is left empty rather than unset. That preserves the current behaviour, since the old code always set the options and so trusted nothing. Returning null would instead let Vert.x fall back to the JVM default trust store and silently trust public CAs.

---

 - use it directly by updating your project with version: `9.7.0-feat-CJ-4304-otel-exporter-base64-keystore-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.7.0-feat-CJ-4304-otel-exporter-base64-keystore-SNAPSHOT/gravitee-node-9.7.0-feat-CJ-4304-otel-exporter-base64-keystore-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/exporter/OTelSslOptionsFactory.java
+++ b/gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/exporter/OTelSslOptionsFactory.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright © 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.opentelemetry.exporter;
+
+import io.gravitee.node.opentelemetry.configuration.OpenTelemetryConfiguration;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.net.JksOptions;
+import io.vertx.core.net.KeyCertOptions;
+import io.vertx.core.net.KeyStoreOptionsBase;
+import io.vertx.core.net.PemKeyCertOptions;
+import io.vertx.core.net.PemTrustOptions;
+import io.vertx.core.net.PfxOptions;
+import io.vertx.core.net.TrustOptions;
+import java.util.Base64;
+import java.util.List;
+import lombok.CustomLog;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Builds Vert.x TLS trust and key-cert options for the OpenTelemetry exporter.
+ *
+ * <p>A jks or pkcs12 store is supplied either as a file path or as Base64-encoded content, the path winning
+ * when both are set. PEM is path-only.
+ *
+ * @author GraviteeSource Team
+ */
+@CustomLog
+@RequiredArgsConstructor
+final class OTelSslOptionsFactory {
+
+    private static final String KEYSTORE_FORMAT_JKS = "JKS";
+    private static final String KEYSTORE_FORMAT_PEM = "PEM";
+    private static final String KEYSTORE_FORMAT_PKCS12 = "PKCS12";
+    private static final List<String> KNOWN_FORMATS = List.of(KEYSTORE_FORMAT_JKS, KEYSTORE_FORMAT_PEM, KEYSTORE_FORMAT_PKCS12);
+
+    private final OpenTelemetryConfiguration configuration;
+
+    KeyCertOptions buildKeyCertOptions() {
+        var type = configuration.getKeystoreType();
+        if (type == null) {
+            return null;
+        }
+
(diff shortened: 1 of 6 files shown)


gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/exporter/OTelSslOptionsFactory.java
```diff
@@ -0,0 +1,152 @@
...
+final class OTelSslOptionsFactory {
+
+    private static final String KEYSTORE_FORMAT_JKS = "JKS";
+    private static final String KEYSTORE_FORMAT_PEM = "PEM";
+    private static final String KEYSTORE_FORMAT_PKCS12 = "PKCS12";
+    private static final List<String> KNOWN_FORMATS = List.of(KEYSTORE_FORMAT_JKS, KEYSTORE_FORMAT_PEM, KEYSTORE_FORMAT_PKCS12);
+
+    private final OpenTelemetryConfiguration configuration;
+
+    KeyCertOptions buildKeyCertOptions() {
```
remibaptistegio, line 51: Would it be possible to reuse the existing builder like io.gravitee.node.vertx.client.ssl.jks.JKSKeyStore (they already implement the logic you implemented here and you will gain the certificate expiration check) ?

remibaptistegio:
In the description of the PR, you pointed an issue with the Logger since its using a different implementation, no ssl configuration is given to it. It means that if a customer wants to configure ssl for both to export traces and logs, he can't for now. Is there another ticket created to track this issue ?

gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/exporter/OTelSslOptionsFactory.java
josephlomasney: I did try that first, but the expiry check pulls in BouncyCastle and it's only test scope across node, so it fails with NoClassDefFoundError.

I think we could declare bcpkix at compile scope on gravitee-node-certificates, though I think that would land BC on every consumer's runtime classpath. Not sure what else that might affect. What do you think? Happy to switch to the shared classes if you think that would be fine.

josephlomasney:
> In the description of the PR, you pointed an issue with the Logger since its using a different implementation, no ssl configuration is given to it. It means that if a customer wants to configure ssl for both to export traces and logs, he can't for now. Is there another ticket created to track this issue ?

No ticket that I've seen. I only came across it while working out what config we show in Cockpit, where the plan for now is to present TLS as traces-only.

Happy to raise one, though I don't think it belongs under this epic. Probably not the Cloud team either, but not sure who it would sit with. Do you know?

gravitee-node-opentelemetry/src/main/java/io/gravitee/node/opentelemetry/exporter/OTelSslOptionsFactory.java
remibaptistegio: Ok no need for now I guess ;)

remibaptistegio:
> > In the description of the PR, you pointed an issue with the Logger since its using a different implementation, no ssl configuration is given to it. It means that if a customer wants to configure ssl for both to export traces and logs, he can't for now. Is there another ticket created to track this issue ?
> 
> No ticket that I've seen. I only came across it while working out what config we show in Cockpit, where the plan for now is to present TLS as traces-only.
> 
> Happy to raise one, though I don't think it belongs under this epic. Probably not the Cloud team either, but not sure who it would sit with. Do you know?

Yes I think we should raise it, maybe for now you could assign it to the Observability project.

josephlomasney:
> > > In the description of the PR, you pointed an issue with the Logger since its using a different implementation, no ssl configuration is given to it. It means that if a customer wants to configure ssl for both to export traces and logs, he can't for now. Is there another ticket created to track this issue ?
> > 
> > 
> > No ticket that I've seen. I only came across it while working out what config we show in Cockpit, where the plan for now is to present TLS as traces-only.
> > Happy to raise one, though I don't think it belongs under this epic. Probably not the Cloud team either, but not sure who it would sit with. Do you know?
> 
> Yes I think we should raise it, maybe for now you could assign it to the Observability project.

Sounds good, raised it as [OBS-55](https://gravitee.atlassian.net/browse/OBS-55) 

[OBS-55]: https://gravitee.atlassian.net/browse/OBS-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

remibaptistegio (approved):
Maybe while the ticket for the logger is not done, we could just add a line in the README explaining that the `ssl` section is only applied to the tracer, wdyt ?

josephlomasney:
> Maybe while the ticket for the logger is not done, we could just add a line in the README explaining that the `ssl` section is only applied to the tracer, wdyt ?

Yeah good point, I added some details in the readme if want to take a look before I merge

remibaptistegio (approved):

PR #602, opened 2026-08-17
fix: reuse a single NetClient in JettyHttpServerProbe

lgw-gravitee:
Vert.x keeps every client until the Vert.x instance closes, so creating one per health check leaked clients for the node lifetime. Create the client lazily once, reuse it, and close the socket after connect.

https://gravitee.atlassian.net/browse/AM-7517

---

 - use it directly by updating your project with version: `9.6.2-fix-jetty-http-server-mem-leak-SNAPSHOT`
 - download it from Artifactory [here](https://odbxikk7vo-artifactory.services.clever-cloud.com/gravitee-snapshots/io/gravitee/node/gravitee-node/9.6.2-fix-jetty-http-server-mem-leak-SNAPSHOT/gravitee-node-9.6.2-fix-jetty-http-server-mem-leak-SNAPSHOT.zip)

### DIFF
--- a/gravitee-node-jetty/src/test/java/io/gravitee/node/jetty/healthcheck/JettyHttpServerProbeTest.java
+++ b/gravitee-node-jetty/src/test/java/io/gravitee/node/jetty/healthcheck/JettyHttpServerProbeTest.java
@@ -1,0 +1,138 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.node.jetty.healthcheck;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+import io.gravitee.node.api.configuration.Configuration;
+import io.gravitee.node.api.healthcheck.Result;
+import io.vertx.core.Vertx;
+import io.vertx.core.internal.VertxInternal;
+import io.vertx.core.net.NetClient;
+import io.vertx.core.net.NetServer;
+import java.lang.reflect.Field;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/**
+ * @author GraviteeSource Team
+ */
+@ExtendWith(MockitoExtension.class)
+class JettyHttpServerProbeTest {
+
+    @Mock
+    private Configuration configuration;
+
+    private Vertx vertx;
+    private NetServer server;
+    private CountDownLatch serverSideClose;
+
+    @BeforeEach
+    void setUp() {
+        vertx = Vertx.vertx();
+    }
+
+    @AfterEach
+    void tearDown() {
+        if (server != null) {
+            server.close().toCompletionStage().toCompletableFuture().join();
+        }
+        vertx.close().toCompletionStage().toCompletableFuture().join();
+    }
+
+    @Test
+    void shouldBeHealthyWhenServerIsListening() throws Exception {
+        givenConfiguration(startServer());
+        JettyHttpServerProbe cut = new JettyHttpServerProbe(configuration, vertx);
+
+        Result result = cut.check().toCompletableFuture().get(10, TimeUnit.SECONDS);
+
+        assertThat(result.isHealthy()).isTrue();
+    }
+
+    @Test
+    void shouldBeUnhealthyWhenServerIsNotListening() throws Exception {
+        int port = startServer();
+        server.close().toCompletionStage().toCompletableFuture().join();
+        server = null;
+        givenConfiguration(port);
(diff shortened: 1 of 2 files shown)


phiz71 (commented):
Same fix as the one merged for APIM-14921 (gravitee-io/gravitee-api-management#19145, plus the 4.12.x backport #19149), and the approach is right. Three notes, none blocking.

**1. `volatile` is redundant — nit**

`private volatile NetClient client` is only ever read and written inside the `synchronized netClient()` getter, so the monitor already gives the visibility guarantee. A plain field does the job.

**2. The health result now waits for the socket close — improvement**

```java
res.result().close().onComplete(ignore -> promise.complete(Result.healthy()));
```

The probe used to answer as soon as the connection succeeded; it now also waits for the close future. If that future never completes the probe never answers at all, and `NodeHealthCheckThread` waits out its timeout instead of getting a healthy result. Completing first keeps the result independent of the teardown:

```java
promise.complete(Result.healthy());
res.result().close();
```

The risk is small on a loopback socket, but it is a failure path that buys nothing.

**3. `shouldCreateSingleNetClientForAllChecks` asserts the shape of the fix, not the leak — nit**

`verify(mockedVertx, times(1)).createNetClient(...)` does pin the reuse, and it would catch a straight revert to one client per check. What it cannot show is the property the fix exists for — that nothing accumulates — because a mocked `Vertx` has no close future to accumulate in.

The assertion that caught this in APIM counts what Vert.x still holds:

```java
Object closeFuture = ((VertxInternal) vertx).closeFuture();
Field childrenField = closeFuture.getClass().getDeclaredField("children");
childrenField.setAccessible(true);
Map<?, ?> children = (Map<?, ?>) childrenField.get(closeFuture);
// count the keys that are NetClient instances
```

Run one check, record the count, run nine more, assert it has not moved. That fails with `expected: 1L but was: 10L` before the fix and passes after, against a real `Vertx` and a real `NetServer`.

**One trap worth knowing about, already handled here**

Closing the `NetSocket` explicitly is not cosmetic: the old code never closed it, `client.close()` took everything down with it. Reusing the client without adding that close would trade a client leak for a socket leak. This PR does it, so nothing to change — flagging it because the same pattern is still open elsewhere (see below).

**Context**

The root cause is in Vert.x, not here. Since 5.0, `createNetClient` registers the internal client in the strong `children` map of its owner's close future, and nothing ever removes it: `NetClientImpl` is not a `NestedCloseable`, so `close()` only tears the connections down. Measured with a standalone loop of 10 000 create + close cycles: 10 000 clients retained on 5.0.12 and on 5.1.6, and 0 on 5.2.0-SNAPSHOT, where `6a0a037e2` ("Introduce context resource cleanup") fixed it upstream. So a future Vert.x upgrade makes this moot, but nothing on the 5.0 or 5.1 lines does.

Two more probes still carry the same pattern, if you want to sweep them in the same pass:

- `GraviteeSpecGenProbe` in `gravitee-spec-gen-agent`
- `KafkaServerProbe.checkUsingTcpProtocol()` in `gravitee-reactor-native-kafka` — this one over TLS, so each retained client also holds its own trust store copy, roughly five times heavier per instance

gravitee-node-jetty/src/test/java/io/gravitee/node/jetty/healthcheck/JettyHttpServerProbeTest.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private void givenConfiguration(int port) {
+        when(configuration.getProperty("jetty.port", Integer.class, 8093)).thenReturn(port);
+        when(configuration.getProperty("jetty.host", "localhost")).thenReturn("localhost");
+    }
+
+    private int startServer() {
+        server =
+            vertx.createNetServer().connectHandler(socket -> {}).listen(0, "localhost").toCompletionStage().toCompletableFuture().join();
```
phiz71: **improvement** — nothing here can observe the socket close.

`connectHandler(socket -> {})` drops the socket, so no test covers the `res.result().close()` on line 66 of the probe. Delete that line and all three tests still pass — yet it is the line that keeps this fix from trading a client leak for a socket leak. The old code never closed the socket either; `client.close()` took it down with the client. Now that the client is shared, that safety net is gone.

Observing it server-side is enough:

```java
private CountDownLatch serverSideClose;

private int startServer() {
    serverSideClose = new CountDownLatch(1);
    server = vertx
        .createNetServer()
        .connectHandler(socket -> socket.closeHandler(event -> serverSideClose.countDown()))
        .listen(0, "localhost")
        .toCompletionStage()
        .toCompletableFuture()
        .join();
    return server.actualPort();
}
```

then one test that runs a check and asserts `serverSideClose.await(10, TimeUnit.SECONDS)` is true.

gravitee-node-jetty/src/test/java/io/gravitee/node/jetty/healthcheck/JettyHttpServerProbeTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+        assertThat(retainedNetClients()).isEqualTo(retainedAfterFirstCheck);
+    }
+
+    private long retainedNetClients() throws Exception {
+        Object closeFuture = ((VertxInternal) vertx).closeFuture();
+        Field childrenField = closeFuture.getClass().getDeclaredField("children");
+        childrenField.setAccessible(true);
+        Map<?, ?> children = (Map<?, ?>) childrenField.get(closeFuture);
+        return children.keySet().stream().filter(NetClient.class::isInstance).count();
```
phiz71: **nit** — this dereferences `children` without a null check.

Vert.x creates the map lazily in `CloseFuture.add` and sets it back to `null` in `cascadeClose`, so it is null until something registers. Not reachable as written, since every caller runs a check first, but it makes the helper unsafe to reuse or to call from a test that does not connect:

```java
return children == null ? 0 : children.keySet().stream().filter(NetClient.class::isInstance).count();
```

A line of comment above the reflection would also earn its keep: it is there because the retention is not observable through any public Vert.x API, which is the whole reason this test exists.

phiz71 (commented):
All three points from the previous round are addressed — `volatile` gone, the promise completed before the socket close, and the mocked-`Vertx` assertion replaced by a real retention count. `shouldNotRetainANetClientPerCheck` now measures the thing the fix is for.

Two notes on the new code, both inline. Neither blocks.

phiz71 (approved):
Both points are addressed: the socket close is now covered by `shouldCloseSocketAfterCheck`, and `retainedNetClients()` guards the null map. The suite pins the three properties that matter — reachable/unreachable, socket released, nothing retained per check.

Nothing further from me.

